PR #16, opened 2017-11-21
Remove build config that is not needed anymore

pekam:
The same result is now implemented in the parent-pom via https://github.com/vaadin/flow-component-base/pull/26

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid-flow/16)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -81,18 +81,4 @@
             <scope>test</scope>
         </dependency>
     </dependencies>
-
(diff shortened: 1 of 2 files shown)


pekam:
It's still an open question how this will be handled, since this apparently breaks every component that doesn't have the resource-folder specified here. So closed for now.

PR #21, opened 2017-11-22
Return method reference for descending Column::getComparator

heruan:
A small change for a great gain: since method references are traceable via reflection, persistence frameworks can obtain from them the original comparator (which may be a custom implementation) set on a column if Grid/Column use only method references to make them serializable.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid-flow/21)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,9 @@
+# The default behavior
+* text=auto
+
+*.java text
+*.xml text
+*.md text
+
+.gitattributes text
+.gitignore text
--- a/pom.xml
+++ b/pom.xml
@@ -18,10 +18,6 @@
 
     <repositories>
         <repository>
-            <id>flow</id>
-            <url>https://repo.vaadin.com/nexus/content/repositories/flow/</url>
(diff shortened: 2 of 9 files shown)


heruan:
See also https://github.com/vaadin/flow/issues/2991.

heruan:
Still, `Grid::createSortingComparator` uses `(x, y) -> 0` as identity for reduction, and this breaks the traceability of the original comparators. Also, this would issue an unnecessary sorting if `sortOrder` is empty.

Why not return an `Optional`, reducing without the identity?

caalador:
Reviewed 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid-flow/21)*

caalador:
Reviewed 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid-flow/21)*

PR #50, opened 2017-12-22
Output sources and javadocs during build

ahie:
Addresses vaadin/flow#3206

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid-flow/50)


ahie:
@manolo yes, moved the configurations into `flow-component-base`.

PR #51, opened 2017-12-22
Fix Grid to send item details data only when the details are actually open

gilberto-torrezan:
In my tests there weren't any duplicated data as the ticket suggest - but a high load of unnecessary data sent from the server. That data is related to the item details - basically Grid was sending the data (and the components) related to item details for every item in the Grid, even when the details weren't shown.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid-flow/51)


ahie:
:lgtm:

---

Reviewed 8 of 8 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid-flow/51#-:-L1M6cO0TtcLyXNEnEtr:bnfp4nl)*

PR #60, opened 2018-01-16
Update generated components and add lumo style to demos

ZheSun88:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid-flow/60)

PR #61, opened 2018-01-17
Fix multiselection bug

pekam:
There was a bug with grid in multiselection mode: The first click on a checkbox in the selection column would do the selection but the checkbox would stay unchecked. The second click would make the checkbox checked, but wouldn't change the selection. The same with unselecting.
This patch fixes the inconsistent state between checkboxes and selected items.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid-flow/61)


pekam:
Decided to put this change on top of https://github.com/vaadin/vaadin-grid-flow/pull/60 since it's related to updating the component and both changes depend on each other.

PR #72, opened 2018-01-25
Remove the support for renderers on headers and footers

gilberto-torrezan:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid-flow/72)

### DIFF
--- a/src/main/java/com/vaadin/flow/component/grid/AbstractColumn.java
+++ b/src/main/java/com/vaadin/flow/component/grid/AbstractColumn.java
@@ -18,8 +18,8 @@ package com.vaadin.flow.component.grid;
 import com.vaadin.flow.component.Component;
 import com.vaadin.flow.component.HasStyle;
 import com.vaadin.flow.component.Synchronize;
-import com.vaadin.flow.data.provider.DataGenerator;
 import com.vaadin.flow.dom.Element;
+import com.vaadin.flow.internal.HtmlUtils;
 import com.vaadin.flow.renderer.ComponentTemplateRenderer;
 import com.vaadin.flow.renderer.TemplateRenderer;
 
@@ -82,7 +82,28 @@ public class AbstractColumn<T extends AbstractColumn<T>> extends Component
     }
 
     @Override
-    public T setHeader(TemplateRenderer<?> renderer) {
+    public T setHeader(String labelText) {
+        return setHeader(TemplateRenderer.of(HtmlUtils.escape(labelText)));
+    }
+
+    @Override
+    public T setFooter(String labelText) {
+        return setFooter(TemplateRenderer.of(HtmlUtils.escape(labelText)));
+    }
+
+    @Override
+    public T setHeader(Component headerComponent) {
+        return setHeader(
+                new ComponentTemplateRenderer<>(() -> headerComponent));
+    }
+
+    @Override
+    public T setFooter(Component footerComponent) {
+        return setFooter(
+                new ComponentTemplateRenderer<>(() -> footerComponent));
+    }
+
+    private T setHeader(TemplateRenderer<?> renderer) {
         if (headerTemplate == null) {
             headerTemplate = new Element("template").setAttribute("class",
                     "header");
@@ -93,8 +114,7 @@ public class AbstractColumn<T extends AbstractColumn<T>> extends Component
         return (T) this;
     }
 
-    @Override
-    public T setFooter(TemplateRenderer<?> renderer) {
+    private T setFooter(TemplateRenderer<?> renderer) {
         if (footerTemplate == null) {
             footerTemplate = new Element("template").setAttribute("class",
                     "footer");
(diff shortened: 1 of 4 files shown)


denis-anisimov:
Reviewed 4 of 4 files at r1.

---

*[src/main/java/com/vaadin/flow/component/grid/ColumnBase.java, line 97 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid-flow/72#-L3l-HyYpGGhygqmClha:-L3l-HyYpGGhygqmClhb:b4r0wd1) ([raw file](https://github.com/vaadin/vaadin-grid-flow/blob/67574937bd118d31de491a23d03f4ca516a8653a/src/main/java/com/vaadin/flow/component/grid/ColumnBase.java#L97)):*
> T setHeader(Component headerComponent);

What does happen with the text that I've set via `setHeader(String)` if I use this method?
Is it true that those two methods are dependent and one cancels another?

If it's so then this is very bad API.
My personal opinion: there should not be two methods which does one thing.
It's very confusing.

What I definitely know: I don't expect this behavior.
So at least I would expect it be written int he Javadoc of both methods.
But still I would reconsider the API since no one reads javadocs.
Though normally I ignored in those questions.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid-flow/72)*

gilberto-torrezan:

---

*[src/main/java/com/vaadin/flow/component/grid/ColumnBase.java, line 97 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid-flow/72#-L3l-HyYpGGhygqmClha:-L3laISskJzL79veenLD:bo9wdvx) ([raw file](https://github.com/vaadin/vaadin-grid-flow/blob/67574937bd118d31de491a23d03f4ca516a8653a/src/main/java/com/vaadin/flow/component/grid/ColumnBase.java#L97)):*

> T setHeader(Component headerComponent);

What does happen with the text that I've set via `setHeader(String)` if I use this method?
Is it true that those two methods are dependent and one cancels another?

If it's so then this is very bad API.
My personal opinion: there should not be two methods which does one thing.
It's very confusing.

What I definitely know: I don't expect this behavior.
So at least I would expect it be written int he Javadoc of both methods.
But still I would reconsider the API since no one reads javadocs.
Though normally I ignored in those questions.
</blockquote></details>

They are meant to be overloads of each other. Just like we had the `setHeader(String)` and `setHeader(TemplateRenderer)` before, but now the renderer support was removed (and replaced for just plain `Component`). So yes, one cancel each other if you use both for some reason.

The sad part is that we can't provide a default implementation anymore, since creating a Component just to render a text would have a negative performance impact for no reason.

Any suggestions on how to improve this?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid-flow/72)*

gilberto-torrezan:
Review status: 3 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/main/java/com/vaadin/flow/component/grid/ColumnBase.java, line 97 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid-flow/72#-L3l-HyYpGGhygqmClha:-L3mGGceqS7EYkz2ZpJq:b-gv1wds) ([raw file](https://github.com/vaadin/vaadin-grid-flow/blob/67574937bd118d31de491a23d03f4ca516a8653a/src/main/java/com/vaadin/flow/component/grid/ColumnBase.java#L97)):*

There is already  the `Text` component which being set via `setHeader(Component)` should do the same as `setText(String)`. The only issue is that the `Text` is not discoverable.
But that could be added into javadocs of the method.
Here and for the footer.

Well, up to you.
What definitely has to be done : javadocs have to be updated.
</blockquote></details>

The performance hit I'm mentioning is that when we use a Component, a `flow-component-renderer` is created to allow Flow to access and modify the component inside a `<template>`. And all that sorcery is useless when just setting text (the text can be inserted directly into the `<template>` without problems). That's why, for the performance point of view, we shouldn't just add a `new Text(string)` here.

Javadocs updated.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid-flow/72)*

denis-anisimov:
Reviewed 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[src/main/java/com/vaadin/flow/component/grid/ColumnBase.java, line 97 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid-flow/72#-L3l-HyYpGGhygqmClha:-L3mQjWkMTq-Yc8lYpC2:bzicr0) ([raw file](https://github.com/vaadin/vaadin-grid-flow/blob/67574937bd118d31de491a23d03f4ca516a8653a/src/main/java/com/vaadin/flow/component/grid/ColumnBase.java#L97)):*

The performance hit I'm mentioning is that when we use a Component, a `flow-component-renderer` is created to allow Flow to access and modify the component inside a `<template>`. And all that sorcery is useless when just setting text (the text can be inserted directly into the `<template>` without problems). That's why, for the performance point of view, we shouldn't just add a `new Text(string)` here.

Javadocs updated.
</blockquote></details>

You don't have to use the same impl in this case.
API can be the same but for performance reasons you may add a special case into the impl which checks whether the `Element` of the `Component` is the text element and apply totally different path in this case.

But OK. As I said it's up to you.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid-flow/72)*

PR #92, opened 2018-02-14
Reuse templates for header, footer and details

pekam:
Instead of always creating new `<template>` elements, reuse the old one if such is already created.

In addition to increasing performance, this also allows changing the header and footer after once set, because previously it would just have added new templates next to the existing ones, and the existing template would still have been used for the header/footer.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid-flow/92)


pekam:
Waiting for https://github.com/vaadin/flow/pull/3555

gilberto-torrezan:
:lgtm:

---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid-flow/92#-:-L5Nj28Sxz8_OdC50xkF:bnfp4nl)*

PR #97, opened 2018-02-20
Make setPageSize public

gilberto-torrezan:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid-flow/97)

### DIFF
--- a/src/main/java/com/vaadin/flow/component/grid/Grid.java
+++ b/src/main/java/com/vaadin/flow/component/grid/Grid.java
@@ -996,7 +996,8 @@ public class Grid<T> extends Component implements HasDataProvider<T>, HasStyle,
     }
 
     /**
-     * Gets the current page size.
+     * Gets the current page size, which is the number of items fetched at a
+     * time from the dataprovider.
      *
      * @return the current page size
      */
@@ -1005,16 +1006,21 @@ public class Grid<T> extends Component implements HasDataProvider<T>, HasStyle,
     }
 
     /**
-     * Sets the page size.
+     * Sets the page size, which is the number of items fetched at a time from
+     * the dataprovider.
      * <p>
-     * This method is private at the moment because the Grid doesn't support
-     * changing the pageSize after the initial load.
+     * Note: the number of items in the server-side memory can be considerably
+     * higher than the page size, since the component can show more than one
+     * page at a time.
+     * <p>
+     * Setting the pageSize after the Grid has been rendered effectively resets
+     * the component, and the current page(s) and sent over again.
      *
      * @param pageSize
      *            the maximum number of items sent per request. Should be
      *            greater than zero
      */
-    private void setPageSize(int pageSize) {
+    public void setPageSize(int pageSize) {
         if (pageSize <= 0) {
             throw new IllegalArgumentException(
                     "The pageSize should be greater than zero. Was "
@@ -1026,7 +1032,9 @@ public class Grid<T> extends Component implements HasDataProvider<T>, HasStyle,
         // https://github.com/vaadin/vaadin-grid/issues/1200
         getElement().setProperty("size", 0);
 
-        getDataCommunicator().setRequestedRange(0, pageSize);
+        getElement().callFunction("$connector.reset");
+        setRequestedRange(0, pageSize);
+        getDataCommunicator().reset();
     }
 
     /**
--- a/src/main/resources/META-INF/resources/frontend/gridConnector.js
+++ b/src/main/resources/META-INF/resources/frontend/gridConnector.js
@@ -237,6 +237,9 @@ window.gridConnector = {
     };
 
     grid.$connector.clear = function(index, length) {
+      if (Object.keys(cache).length === 0){
+        return;
+      }
       if (index % grid.pageSize != 0) {
         throw 'Got cleared data for index ' + index + ' which is not aligned with the page size of ' + grid.pageSize;
       }
@@ -257,6 +260,21 @@ window.gridConnector = {
         updateGridCache(page);
       }
     };
+
+    grid.$connector.reset = function() {
+      grid.size = 0;
+      deleteObjectContents(cache);
+      deleteObjectContents(grid._cache.items);
+      lastRequestedRange = [0, 0];
+      grid._assignModels();
+    };
+
+    const deleteObjectContents = function(obj) {
+      let props = Object.keys(obj);
+      for (let i = 0; i < props.length; i++) {
+        delete obj[props[i]];
+      }
+    }
 
     grid.$connector.updateSize = function(newSize) {
       grid.size = newSize;
--- a/src/test/java/com/vaadin/flow/component/grid/it/GridPageSizePage.java
+++ b/src/test/java/com/vaadin/flow/component/grid/it/GridPageSizePage.java
@@ -19,7 +19,9 @@ import java.util.stream.IntStream;
 
 import com.vaadin.flow.component.grid.Grid;
 import com.vaadin.flow.component.html.Div;
+import com.vaadin.flow.component.html.Input;
 import com.vaadin.flow.component.html.Label;
+import com.vaadin.flow.component.html.NativeButton;
 import com.vaadin.flow.data.provider.DataProvider;
 import com.vaadin.flow.data.provider.Query;
 import com.vaadin.flow.router.Route;
@@ -49,7 +51,15 @@ public class GridPageSizePage extends Div {
         info = new Label();
         info.setId("query-info");
 
-        add(grid, info);
+        Input size = new Input();
+        size.setId("size-input");
+        NativeButton button = new NativeButton("Change page size", event -> {
+            int pageSize = Integer.parseInt(size.getValue());
+            grid.setPageSize(pageSize);
+        });
+        button.setId("size-submit");
+
+        add(grid, info, new Div(size, button));
     }
 
     private IntStream getStream(Query<String, Void> query) {
--- a/src/test/java/com/vaadin/flow/component/grid/it/GridPageSizePageIT.java
+++ b/src/test/java/com/vaadin/flow/component/grid/it/GridPageSizePageIT.java
@@ -15,7 +15,10 @@
  */
 package com.vaadin.flow.component.grid.it;
 
+import java.util.List;
+
 import org.junit.Assert;
+import org.junit.Before;
 import org.junit.Test;
 import org.openqa.selenium.By;
 import org.openqa.selenium.WebElement;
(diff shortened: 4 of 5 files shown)


denis-anisimov:
Reviewed 4 of 4 files at r1.
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[src/main/resources/META-INF/resources/frontend/gridConnector.js, line 119 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid-flow/97#-L5mferL2qeswOJspN1i:-L5mferL2qeswOJspN1j:b-uhizly) ([raw file](https://github.com/vaadin/vaadin-grid-flow/blob/9819e8cf4332535e18ddbafa5b73f0ad3c76631d/src/main/resources/META-INF/resources/frontend/gridConnector.js#L119)):*
> let first = Math.max(0,  firstNeededPage - extraPageBuffer);

So it has been a window with extra data below `firstNeededPage` and above `lastNeededPage`.
Now the extra data is just above `lastNeededPage`. 
Don't really know how the `first` is used (may be it's OK) but it became asymmetrical. 

---

*[src/test/java/com/vaadin/flow/component/grid/it/GridPageSizePageIT.java, line 31 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid-flow/97#-L5mgdCTwB8KlEEcV9GA:-L5mgdCTwB8KlEEcV9GB:b-w4mysy) ([raw file](https://github.com/vaadin/vaadin-grid-flow/blob/2b8dd3ed95646cd05851e2b78f0c1962662ece98/src/test/java/com/vaadin/flow/component/grid/it/GridPageSizePageIT.java#L31)):*
> GridPageSizePageIT

The test doesn't check whether the Grid is functional.
It just checks the `pageSize` property of the `Grid`.
It might be that the property value is correct but a `Grid` doesn't show anything.......
May be there should be some checks against the visual content of the `Grid`, some smoke tests for scrolling?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid-flow/97)*

gilberto-torrezan:
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[src/main/resources/META-INF/resources/frontend/gridConnector.js, line 119 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid-flow/97#-L5mferL2qeswOJspN1i:-L5mvDGr4Grdcu8kLuPF:b9v0rxe) ([raw file](https://github.com/vaadin/vaadin-grid-flow/blob/9819e8cf4332535e18ddbafa5b73f0ad3c76631d/src/main/resources/META-INF/resources/frontend/gridConnector.js#L119)):*

> let first = Math.max(0,  firstNeededPage - extraPageBuffer);

So it has been a window with extra data below `firstNeededPage` and above `lastNeededPage`.
Now the extra data is just above `lastNeededPage`. 
Don't really know how the `first` is used (may be it's OK) but it became asymmetrical. 
</blockquote></details>

In my tests the `grid._virtualStart` always covered what was needed by the Grid. Bringing one extra page (from the start) turned out to be unnecessary.

But about the `gird_virtualEnd`, on the other hand, is a different case. We need to inform the server something beyond the current range in the Grid, otherwise the server will remove pages that are still in use by the Grid.

---

*[src/test/java/com/vaadin/flow/component/grid/it/GridPageSizePageIT.java, line 31 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid-flow/97#-L5mgdCTwB8KlEEcV9GA:-L5mwG1BnQVbwxx3kKrE:b-b84ai7) ([raw file](https://github.com/vaadin/vaadin-grid-flow/blob/2b8dd3ed95646cd05851e2b78f0c1962662ece98/src/test/java/com/vaadin/flow/component/grid/it/GridPageSizePageIT.java#L31)):*

> GridPageSizePageIT

The test doesn't check whether the Grid is functional.
It just checks the `pageSize` property of the `Grid`.
It might be that the property value is correct but a `Grid` doesn't show anything.......
May be there should be some checks against the visual content of the `Grid`, some smoke tests for scrolling?
</blockquote></details>

Smoke tests maybe. Concrete exact tests are tricky. Since the exact amount of data depends on the size and scroll position of the Grid, and the window size of the IT is unpredictable, making a precise test for it is very unstable.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid-flow/97)*

Legioth:
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[src/main/resources/META-INF/resources/frontend/gridConnector.js, line 119 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid-flow/97#-L5mferL2qeswOJspN1i:-L5n8JVCCxf8zB374DBZ:b4neyq) ([raw file](https://github.com/vaadin/vaadin-grid-flow/blob/9819e8cf4332535e18ddbafa5b73f0ad3c76631d/src/main/resources/META-INF/resources/frontend/gridConnector.js#L119)):*

In my tests the `grid._virtualStart` always covered what was needed by the Grid. Bringing one extra page (from the start) turned out to be unnecessary.

But about the `gird_virtualEnd`, on the other hand, is a different case. We need to inform the server something beyond the current range in the Grid, otherwise the server will remove pages that are still in use by the Grid.
</blockquote></details>

So what you're saying is that scrolling up should be feel slow because nothing is seen until data has been fetched, whereas scrolling own would be smooth as data has been fetched proactively before it's actually needed?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid-flow/97)*

Legioth:
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[src/main/resources/META-INF/resources/frontend/gridConnector.js, line 119 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid-flow/97#-L5mferL2qeswOJspN1i:-L5n8jbsJw7huXt5ETYL:bnlkyhg) ([raw file](https://github.com/vaadin/vaadin-grid-flow/blob/9819e8cf4332535e18ddbafa5b73f0ad3c76631d/src/main/resources/META-INF/resources/frontend/gridConnector.js#L119)):*
<details><summary><i>Previously, Legioth (Leif Åstrand) wrote…</i></summary><blockquote>

So what you're saying is that scrolling up should be feel slow because nothing is seen until data has been fetched, whereas scrolling own would be smooth as data has been fetched proactively before it's actually needed?
</blockquote></details>

And maybe a better question: What does this change have to do with making it possible to change the page size?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid-flow/97)*

gilberto-torrezan:
Review status: 3 of 5 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[src/main/resources/META-INF/resources/frontend/gridConnector.js, line 119 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid-flow/97#-L5mferL2qeswOJspN1i:-L5nHQJiikv5rEhsqz6t:b-l53hgj) ([raw file](https://github.com/vaadin/vaadin-grid-flow/blob/9819e8cf4332535e18ddbafa5b73f0ad3c76631d/src/main/resources/META-INF/resources/frontend/gridConnector.js#L119)):*
<details><summary><i>Previously, Legioth (Leif Åstrand) wrote…</i></summary><blockquote>

And maybe a better question: What does this change have to do with making it possible to change the page size?
</blockquote></details>

The actual problem to be fixed is to reduce memory consumption. That's all this is about. Being able to set the pageSize is a thing that can mitigate the problem, but there are  other things that could help as well.

The first thing I tried was to remove the `extraPageBuffer` var. To make it work properly, it would require substantial changes either on the client and on the server (see Slack). So to avoid breaking things I reduced it from 2 to 1.

Another thing was to deal with the ranges. As I said, I didn't find any issues in reducing the initial range, And that will save one page of items per Grid on the server. Sounds worth it.

We can revert it back based on user feedback if necessary, since it's a small change. Or, if needed, we can export a configuration option to set those values in the future.

---

*[src/test/java/com/vaadin/flow/component/grid/it/GridPageSizePageIT.java, line 31 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid-flow/97#-L5mgdCTwB8KlEEcV9GA:-L5nIv4-33_jrYV8o8MM:bjne8e2) ([raw file](https://github.com/vaadin/vaadin-grid-flow/blob/2b8dd3ed95646cd05851e2b78f0c1962662ece98/src/test/java/com/vaadin/flow/component/grid/it/GridPageSizePageIT.java#L31)):*

Smoke tests maybe. Concrete exact tests are tricky. Since the exact amount of data depends on the size and scroll position of the Grid, and the window size of the IT is unpredictable, making a precise test for it is very unstable.
</blockquote></details>

Added a smoke test to read the cell contents after the `pageSize` is set.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid-flow/97)*

gilberto-torrezan:
Review status: 3 of 5 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/vaadin-grid-flow/97#-L5nJMg9G3i4bgPAHVoO:-L5nJMg9G3i4bgPAHVoP:b-w8hxfu) (no related file):*
The test which is failing also fails on my machine, but passes on @caalador 's. No idea what's going on - it also fails on master (so it's not related to this patch). Investigating...

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid-flow/97)*

denis-anisimov:
Reviewed 2 of 2 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid-flow/97)*

Legioth:
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[src/main/resources/META-INF/resources/frontend/gridConnector.js, line 119 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid-flow/97#-L5mferL2qeswOJspN1i:-L5nP6SUpdKv4_ERvDQu:b-9ten4x) ([raw file](https://github.com/vaadin/vaadin-grid-flow/blob/9819e8cf4332535e18ddbafa5b73f0ad3c76631d/src/main/resources/META-INF/resources/frontend/gridConnector.js#L119)):*

The actual problem to be fixed is to reduce memory consumption. That's all this is about. Being able to set the pageSize is a thing that can mitigate the problem, but there are  other things that could help as well.

The first thing I tried was to remove the `extraPageBuffer` var. To make it work properly, it would require substantial changes either on the client and on the server (see Slack). So to avoid breaking things I reduced it from 2 to 1.

Another thing was to deal with the ranges. As I said, I didn't find any issues in reducing the initial range, And that will save one page of items per Grid on the server. Sounds worth it.

We can revert it back based on user feedback if necessary, since it's a small change. Or, if needed, we can export a configuration option to set those values in the future.
</blockquote></details>

I would prefer to have only one functional change per commit / PR.

Also, I think the discussion about memory consumption vs awesome scrolling UX requires slightly more nuance than a hardcoded value, regardless of whether it's 0, 1 or 2.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid-flow/97)*

gilberto-torrezan:
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[src/main/resources/META-INF/resources/frontend/gridConnector.js, line 119 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid-flow/97#-L5mferL2qeswOJspN1i:-L5nREFhOCzZlytszmkR:bduqovw) ([raw file](https://github.com/vaadin/vaadin-grid-flow/blob/9819e8cf4332535e18ddbafa5b73f0ad3c76631d/src/main/resources/META-INF/resources/frontend/gridConnector.js#L119)):*
<details><summary><i>Previously, Legioth (Leif Åstrand) wrote…</i></summary><blockquote>

I would prefer to have only one functional change per commit / PR.

Also, I think the discussion about memory consumption vs awesome scrolling UX requires slightly more nuance than a hardcoded value, regardless of whether it's 0, 1 or 2.
</blockquote></details>

Agreed. Sounds like a nice idea to discuss after the beta blockers.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid-flow/97)*

gilberto-torrezan:
Review status: 4 of 5 files reviewed at latest revision, 2 unresolved discussions.

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/vaadin-grid-flow/97#-L5nJMg9G3i4bgPAHVoO:-L5nhHNO1OecbiegUeFT:b73q5w4) (no related file):*

The test which is failing also fails on my machine, but passes on @caalador 's. No idea what's going on - it also fails on master (so it's not related to this patch). Investigating...
</blockquote></details>

Needs to merge this https://github.com/vaadin/vaadin-grid-flow/pull/98 to fix the broken test.

---

*[src/main/resources/META-INF/resources/frontend/gridConnector.js, line 119 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid-flow/97#-L5mferL2qeswOJspN1i:-L5nh6iyU5jRDi06KEUs:b-p4qio) ([raw file](https://github.com/vaadin/vaadin-grid-flow/blob/9819e8cf4332535e18ddbafa5b73f0ad3c76631d/src/main/resources/META-INF/resources/frontend/gridConnector.js#L119)):*

Agreed. Sounds like a nice idea to discuss after the beta blockers.
</blockquote></details>

Created a separate PR just for the `extraPageBuffer` thing: https://github.com/vaadin/vaadin-grid-flow/pull/99

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid-flow/97)*

denis-anisimov:
Reviewed 1 of 1 files at r3.
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid-flow/97)*

denis-anisimov:
Reviewed 1 of 1 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid-flow/97)*

PR #99, opened 2018-02-20
Refactor the logic behind the extraPageBuffer to consider visible rows instead

gilberto-torrezan:
This will reduce the server-side memory consumption, since less pages will be requested at a time.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid-flow/99)

### DIFF
--- a/src/main/resources/META-INF/resources/frontend/gridConnector.js
+++ b/src/main/resources/META-INF/resources/frontend/gridConnector.js
@@ -1,11 +1,10 @@
 window.gridConnector = {
-  initLazy: function(grid) {
-    
+  initLazy: function(grid) {    
     // Check whether the connector was already initialized for the grid
     if (grid.$connector){
       return;
     }
-    const extraPageBuffer = 1;
+    const visibleCountMultiplier = 1;
     const pageCallbacks = {};
     const cache = {};
     let lastRequestedRange = [0, 0];
@@ -118,15 +117,20 @@ window.gridConnector = {
       }
       // Determine what to fetch based on scroll position and not only
       // what grid asked for
-      let firstNeededPage = Math.min(page, grid._getPageForIndex(grid._virtualStart + grid._vidxOffset));
-      let lastNeededPage = Math.max(page, grid._getPageForIndex(grid._virtualEnd + grid._vidxOffset));
+      let visibleCount = grid._virtualEnd - grid._virtualStart;
+      let buffer = visibleCount * visibleCountMultiplier;
+      let firstNeededIndex = Math.max(0, grid._virtualStart + grid._vidxOffset - buffer);
+      let lastNeededIndex = Math.min(grid._virtualEnd + grid._vidxOffset + buffer, grid.size);
+
+      let firstNeededPage = Math.min(page, grid._getPageForIndex(firstNeededIndex));
+      let lastNeededPage = Math.max(page, grid._getPageForIndex(lastNeededIndex));
 
       let first = Math.max(0,  firstNeededPage);
-      let last = Math.min(lastNeededPage + extraPageBuffer, Math.floor(grid.size / grid.pageSize) + 1);
+      let last = Math.min(lastNeededPage, Math.floor(grid.size / grid.pageSize) + 1);
 
       if (lastRequestedRange[0] != first || lastRequestedRange[1] != last) {
         lastRequestedRange = [first, last];
-        let count = Math.max(1, last - first);
+        let count = last - first + 1;
         grid.$server.setRequestedRange(first * grid.pageSize, count * grid.pageSize);
       }
     }
--- a/src/main/resources/META-INF/resources/frontend/gridConnector.js
+++ b/src/main/resources/META-INF/resources/frontend/gridConnector.js
@@ -4,7 +4,6 @@ window.gridConnector = {
     if (grid.$connector){
       return;
     }
-    const visibleCountMultiplier = 1;
     const pageCallbacks = {};
     const cache = {};
     let lastRequestedRange = [0, 0];
@@ -117,8 +116,12 @@ window.gridConnector = {
       }
       // Determine what to fetch based on scroll position and not only
       // what grid asked for
-      let visibleCount = grid._virtualEnd - grid._virtualStart;
-      let buffer = visibleCount * visibleCountMultiplier;
+
+      // The buffer size could be multiplied by some constant defined by the user,
+      // if he needs to reduce the number of items sent to the Grid to improve performance
+      // or to increase it to make Grid smoother when scrolling
+      let buffer = grid._virtualEnd - grid._virtualStart;
+      
       let firstNeededIndex = Math.max(0, grid._virtualStart + grid._vidxOffset - buffer);
       let lastNeededIndex = Math.min(grid._virtualEnd + grid._vidxOffset + buffer, grid.size);
 
--- a/src/main/resources/META-INF/resources/frontend/gridConnector.js
+++ b/src/main/resources/META-INF/resources/frontend/gridConnector.js
@@ -5,7 +5,7 @@ window.gridConnector = {
     if (grid.$connector){
       return;
     }
-    const extraPageBuffer = 2;
+    const extraPageBuffer = 1;
     const pageCallbacks = {};
     const cache = {};
     let lastRequestedRange = [0, 0];
@@ -121,7 +121,7 @@ window.gridConnector = {
       let firstNeededPage = Math.min(page, grid._getPageForIndex(grid._virtualStart + grid._vidxOffset));
       let lastNeededPage = Math.max(page, grid._getPageForIndex(grid._virtualEnd + grid._vidxOffset));
 
-      let first = Math.max(0,  firstNeededPage - extraPageBuffer);
+      let first = Math.max(0,  firstNeededPage);
       let last = Math.min(lastNeededPage + extraPageBuffer, Math.floor(grid.size / grid.pageSize) + 1);
 
       if (lastRequestedRange[0] != first || lastRequestedRange[1] != last) {


Legioth:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/vaadin-grid-flow/99#-L5r_f9BtaCbEKAiSgf0:-L5r_f9BtaCbEKAiSgf1:b-ly1x7q) (no related file):*
This will reduce server-side memory consumption, but it will also have some degree of impact on the user experience when scrolling since it increases the risk that empty rows would have to be rendered before the required data has been loaded from the server.

For scrolling downwards you would probably have to scroll quite quickly (or have a slow connection) to notice the difference since there would still be at least 50 rows worth of buffer (with the default page size setting). When scrolling up on the other hand, there will in some cases be 0 rows of buffer, which means that there will occasionally be empty rows that are only populated after a small delay regardless how slowly you scroll. I suspect the visual impact of this will be extra obvious in cases with variable row heights since the things would jump around a bit to adjust once the actual height becomes known.

I propose applying a buffer based on the number of visible rows before converting to page indices instead the current implementation that applies a constant buffer after converting to pages. The basic formula would be something like this (plus various bounds checks and stuff):

```
visibleCount = lastVisibleIndex - firstVisibleIndex;
buffer = visibleCount * <some suitable multiplier, e.g. 1>

firstNeededIndex = firstVisibleIndex - buffer;
lastNeededIndex = lastVisibleIndex + buffer;

firstNeededPage = getPageForIndex(firstNeededIndex);
lastNeededPage = getPageForIndex(lastNeededIndex);
```

In the Bakery case where `visibleCount` is usually less than 10, this would mean that the total number of items to request before adjusting to page indices would be less than 30 with 1 as the multiplier value. This means that slightly less than half of the scrolling positions would need only one page in the cache (assuming the default page size of 50 is used) and the remainder would lead to two pages in the cache. For a smaller screen with only 5 visible rows, two cached pages would be needed only about 30% of the time. 3 pages at the same time would be needed only if the total number of needed items exceeds 50, i.e if more than 20 items are visible at once.

My suggestion would thus never use more memory for Bakery compared to the current patch that would always keep two pages cached (except when scrolled to the last page), but it would use less memory in about 50% of the cases (if the multiplier is 1) and also remove the risk of temporarily seeing empty rows while scrolling up since those rows would be fetched when the user is ~10 rows from the page boundary instead of when reaching the boundary.

We might even consider bumping the multiplier to 2, which would bring the total number of items in the Bakery case to around 50 and thus have two items in the cache all times except when perfectly aligned to a page boundary, but still never need 3 pages.

For reference, Grid in FW8 uses a flexible multiplier between 3 and 5 (lower limit for requesting more and upper limit for expunging). FW8 doesn't have to align the effective cache to page boundaries, but there would still on average be 90 items cached in the Bakery case with 10 visible items.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid-flow/99)*

gilberto-torrezan:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/vaadin-grid-flow/99#-L5r_f9BtaCbEKAiSgf0:-L5sNK1M0cxTyijCbYE-:bniindy) (no related file):*
<details><summary><i>Previously, Legioth (Leif Åstrand) wrote…</i></summary><blockquote>

This will reduce server-side memory consumption, but it will also have some degree of impact on the user experience when scrolling since it increases the risk that empty rows would have to be rendered before the required data has been loaded from the server.

For scrolling downwards you would probably have to scroll quite quickly (or have a slow connection) to notice the difference since there would still be at least 50 rows worth of buffer (with the default page size setting). When scrolling up on the other hand, there will in some cases be 0 rows of buffer, which means that there will occasionally be empty rows that are only populated after a small delay regardless how slowly you scroll. I suspect the visual impact of this will be extra obvious in cases with variable row heights since the things would jump around a bit to adjust once the actual height becomes known.

I propose applying a buffer based on the number of visible rows before converting to page indices instead the current implementation that applies a constant buffer after converting to pages. The basic formula would be something like this (plus various bounds checks and stuff):

```
visibleCount = lastVisibleIndex - firstVisibleIndex;
buffer = visibleCount * <some suitable multiplier, e.g. 1>

firstNeededIndex = firstVisibleIndex - buffer;
lastNeededIndex = lastVisibleIndex + buffer;

firstNeededPage = getPageForIndex(firstNeededIndex);
lastNeededPage = getPageForIndex(lastNeededIndex);
```

In the Bakery case where `visibleCount` is usually less than 10, this would mean that the total number of items to request before adjusting to page indices would be less than 30 with 1 as the multiplier value. This means that slightly less than half of the scrolling positions would need only one page in the cache (assuming the default page size of 50 is used) and the remainder would lead to two pages in the cache. For a smaller screen with only 5 visible rows, two cached pages would be needed only about 30% of the time. 3 pages at the same time would be needed only if the total number of needed items exceeds 50, i.e if more than 20 items are visible at once.

My suggestion would thus never use more memory for Bakery compared to the current patch that would always keep two pages cached (except when scrolled to the last page), but it would use less memory in about 50% of the cases (if the multiplier is 1) and also remove the risk of temporarily seeing empty rows while scrolling up since those rows would be fetched when the user is ~10 rows from the page boundary instead of when reaching the boundary.

We might even consider bumping the multiplier to 2, which would bring the total number of items in the Bakery case to around 50 and thus have two items in the cache all times except when perfectly aligned to a page boundary, but still never need 3 pages.

For reference, Grid in FW8 uses a flexible multiplier between 3 and 5 (lower limit for requesting more and upper limit for expunging). FW8 doesn't have to align the effective cache to page boundaries, but there would still on average be 90 items cached in the Bakery case with 10 visible items.
</blockquote></details>

Thanks. testing the idea right now.

I'm wondering if the responsiveness of the Grid would affect this. For instance: in a mobile device in the vertical position, the rows could be taller (to better accommodate long text), and then when switching to the horizontal position, the rows could return to normal. These transitions would affect the `visibleCount` between one call and another. The same goes for resizing panels or windows.

I don't know whether Grid is smart enough to handle that by itself though. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid-flow/99)*

Legioth:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/vaadin-grid-flow/99#-L5r_f9BtaCbEKAiSgf0:-L5sh9qBWCtLOBC_e83r:bkkbqf3) (no related file):*

Thanks. testing the idea right now.

I'm wondering if the responsiveness of the Grid would affect this. For instance: in a mobile device in the vertical position, the rows could be taller (to better accommodate long text), and then when switching to the horizontal position, the rows could return to normal. These transitions would affect the `visibleCount` between one call and another. The same goes for resizing panels or windows.

I don't know whether Grid is smart enough to handle that by itself though. 

</blockquote></details>

I guess we might be end up in a slightly sub-optimal situation (i.e. retain cached items that could now be discarded) after a dimension change if it doesn't also trigger Grid to call the `dataProvider` function. Things should however sort out themselves the next time `dataProvider` is run since everything is calculated based on `_virtualEnd` and `_virtualStart` that should be directly based on what's currently shown in the grid.

To make things even smoother, we might see if there would also be some other event that should be used for triggering the a new calculation and potentially send do a `setRequestedRange` RPC. I would still investigate that separately from this fix.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid-flow/99)*

SomeoneToIgnore:
If you don't feel like changing something after reading the comments, merge it.

---

Reviewed 1 of 1 files at r1.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[src/main/resources/META-INF/resources/frontend/gridConnector.js, line 7 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid-flow/99#-L6WjkbhcLOE4TrVYSCm:-L6WjkbhcLOE4TrVYSCn:b-wmh5jd) ([raw file](https://github.com/vaadin/vaadin-grid-flow/blob/86da8aaeae235b54872c26a34021bdbe9c69f454/src/main/resources/META-INF/resources/frontend/gridConnector.js#L7)):*
> ```JavaScript
>       return;
>     }
>     const visibleCountMultiplier = 1;
> ```

Are you able to change it in runtime somehow?
If not (which is true, judging by the `const` modifier), don't really see the benefit of multiplying numbers by one.

If you would like to say something about "later we might change it" then introduce it later also.
Or change it now.

How do you pick this number, is it just an arbitrary guess?
Also, to me it would make sense to have it closer to the actual usage, into the `dataProvider` method.

---

*[src/main/resources/META-INF/resources/frontend/gridConnector.js, line 123 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid-flow/99#-L6WmrI6-3wDKOrd1z-O:-L6WmrI6-3wDKOrd1z-P:b5422rs) ([raw file](https://github.com/vaadin/vaadin-grid-flow/blob/86da8aaeae235b54872c26a34021bdbe9c69f454/src/main/resources/META-INF/resources/frontend/gridConnector.js#L123)):*
> ```JavaScript
>       let buffer = visibleCount * visibleCountMultiplier;
>       let firstNeededIndex = Math.max(0, grid._virtualStart + grid._vidxOffset - buffer);
>       let lastNeededIndex = Math.min(grid._virtualEnd + grid._vidxOffset + buffer, grid.size);
> ```

Not sure if it's ok or not, but it looks like you load 2x`buffer` instead of just one.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid-flow/99#-:-L6Wn8g9vtvHlX5oq42U:b3x8fgx)*

gilberto-torrezan:
Review status: all files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

*[src/main/resources/META-INF/resources/frontend/gridConnector.js, line 7 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid-flow/99#-L6WjkbhcLOE4TrVYSCm:-L6WnpFbBM6PdRgTDSeQ:bktqldd) ([raw file](https://github.com/vaadin/vaadin-grid-flow/blob/86da8aaeae235b54872c26a34021bdbe9c69f454/src/main/resources/META-INF/resources/frontend/gridConnector.js#L7)):*
<details><summary><i>Previously, SomeoneToIgnore (Kirill Bulatov) wrote…</i></summary><blockquote>

Are you able to change it in runtime somehow?
If not (which is true, judging by the `const` modifier), don't really see the benefit of multiplying numbers by one.

If you would like to say something about "later we might change it" then introduce it later also.
Or change it now.

How do you pick this number, is it just an arbitrary guess?
Also, to me it would make sense to have it closer to the actual usage, into the `dataProvider` method.
</blockquote></details>

Nope, it is not changeable at the moment. Sure, I can replace it by a comment instead.

---

*[src/main/resources/META-INF/resources/frontend/gridConnector.js, line 123 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid-flow/99#-L6WmrI6-3wDKOrd1z-O:-L6WneY3X3023x35dYTX:b-ze8ghh) ([raw file](https://github.com/vaadin/vaadin-grid-flow/blob/86da8aaeae235b54872c26a34021bdbe9c69f454/src/main/resources/META-INF/resources/frontend/gridConnector.js#L123)):*
<details><summary><i>Previously, SomeoneToIgnore (Kirill Bulatov) wrote…</i></summary><blockquote>

Not sure if it's ok or not, but it looks like you load 2x`buffer` instead of just one.
</blockquote></details>

Intentional. It is like that to support smooth scrolling up and downwards.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid-flow/99)*

gilberto-torrezan:
Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---

*[src/main/resources/META-INF/resources/frontend/gridConnector.js, line 7 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid-flow/99#-L6WjkbhcLOE4TrVYSCm:-L6WpX76kSUXksZa66Em:b-896fix) ([raw file](https://github.com/vaadin/vaadin-grid-flow/blob/86da8aaeae235b54872c26a34021bdbe9c69f454/src/main/resources/META-INF/resources/frontend/gridConnector.js#L7)):*

Nope, it is not changeable at the moment. Sure, I can replace it by a comment instead.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid-flow/99)*

SomeoneToIgnore:
:lgtm:

---

Reviewed 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid-flow/99#-:-L6Wp_CWlEPHa2uLvYso:bnfp4nl)*

PR #132, opened 2018-03-27
Remove flow-grid-component-renderer

### LINKED ISSUE #123: Broken scroll after applying dataProvider filter
Grid scroll is broken when grid has dataProvider and ComponentRenderer column and some filter is applied to the dataProvider.

Prerequisites:
vaadin-grid (1.0.0.beta3, but it was reproduced earlier as well)
ComponentRenderer (flow 1.0.0.beta2)
Tried browsers:
Chrome: no exceptions
Firefox: `TypeError: a is null`

Simple example: https://github.com/alexberazouski/skeleton-starter-flow-spring/tree/bug/grid-scroll

1. Open localhost:8080
2. Scroll the grid 2-3 page down
3. Check SetFilter checkbox
4. Scroll grid again

See that grid always returns back the scroll to the initial position

Bug is not reproduced if grid doesn't have the ComponentRenderer column.

pekam:
None of the problems that this webcomponent solved can be reproduced
anymore: Cells are resized correctly and old contents are not shown when
scrolling very fast. This file also caused a bug which is simply fixed
by removing.

Fixes #123

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid-flow/132)


pekam:
I tested manually how this affects performance on IE11. Based on the tests we get 10-40% improvements in the initial render time by removing `flow-grid-component-renderer`.

For example, with 40 columns using `ComponentRenderer` the rendering times were around 25 seconds vs 15 seconds.

gilberto-torrezan:
:lgtm:

---

Reviewed 5 of 5 files at r1.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/vaadin/vaadin-grid-flow/132#-:-L8g8iSp7esQigguqoJQ:bnfp4nl)*

PR #176, opened 2018-05-02
Render header components only once in a roundtrip

### LINKED ISSUE #172: Appending multiple HeaderRows with components bug
When you append more than one HeaderRow during the same server roundtrip and use components in the HeaderCells instead of text, the contents of the first appended row remain on the bottom row, as can be seen in this example: https://github.com/vaadin/vaadin-grid-flow/issues/170

The `appendHeaderRow()` implementation moves renderers from `<vaadin-grid-column>`s to `<vaadin-grid-column-group>`s but apparently overriding the rendering still makes the original rendering happen. This is most likely caused by the `ComponentRenderer` registering a `beforeClientResponse` command which sets the template.

pekam:
Instead of applying ComponentRenderer for headers every time the
renderer is changed, apply the current renderer beforeClientResponse so
that only the last renderer is used.

Fix #172

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid-flow/176)

### DIFF
--- a/src/main/java/com/vaadin/flow/component/grid/AbstractColumn.java
+++ b/src/main/java/com/vaadin/flow/component/grid/AbstractColumn.java
@@ -86,12 +86,8 @@ public class AbstractColumn<T extends AbstractColumn<T>> extends Component
 
     protected void setHeaderRenderer(Renderer<?> renderer) {
         headerRenderer = renderer;
-        if (renderer instanceof ComponentRenderer<?, ?>) {
-            getElement().getNode().runWhenAttached(ui -> ui
-                    .beforeClientResponse(this, context -> renderHeader()));
-        } else {
-            renderHeader();
-        }
+        getElement().getNode().runWhenAttached(
+                ui -> ui.beforeClientResponse(this, context -> renderHeader()));
     }
 
     protected Rendering<?> renderHeader() {
@@ -110,12 +106,8 @@ public class AbstractColumn<T extends AbstractColumn<T>> extends Component
 
     protected void setFooterRenderer(Renderer<?> renderer) {
         footerRenderer = renderer;
-        if (renderer instanceof ComponentRenderer<?, ?>) {
-            getElement().getNode().runWhenAttached(ui -> ui
-                    .beforeClientResponse(this, context -> renderFooter()));
-        } else {
-            renderFooter();
-        }
+        getElement().getNode().runWhenAttached(
+                ui -> ui.beforeClientResponse(this, context -> renderFooter()));
     }
 
     protected Rendering<?> renderFooter() {
--- a/src/main/java/com/vaadin/flow/component/grid/AbstractColumn.java
+++ b/src/main/java/com/vaadin/flow/component/grid/AbstractColumn.java
@@ -84,50 +84,63 @@ public class AbstractColumn<T extends AbstractColumn<T>> extends Component
         return !getElement().getProperty("hidden", false);
     }
 
-    protected Rendering<?> renderHeader(Renderer<?> renderer) {
+    protected void setHeaderRenderer(Renderer<?> renderer) {
         headerRenderer = renderer;
+        if (renderer instanceof ComponentRenderer<?, ?>) {
+            getElement().getNode().runWhenAttached(ui -> ui
+                    .beforeClientResponse(this, context -> renderHeader()));
+        } else {
+            renderHeader();
+        }
+    }
+
+    protected Rendering<?> renderHeader() {
         if (headerTemplate != null) {
             headerTemplate.removeFromParent();
             headerTemplate = null;
         }
-        if (renderer == null) {
+        if (headerRenderer == null) {
             return null;
         }
-        Rendering<?> rendering = renderer.render(getElement(), null);
+        Rendering<?> rendering = headerRenderer.render(getElement(), null);
         headerTemplate = rendering.getTemplateElement();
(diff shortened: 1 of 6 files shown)


gilberto-torrezan:
Reviewed 4 of 4 files at r1.

---

*[src/main/java/com/vaadin/flow/component/grid/AbstractColumn.java, line 93 at r1](https://reviewable.io/reviews/vaadin/vaadin-grid-flow/176#-LBZ_oqDCllm-UJvUtyG:-LBZ_oqDCllm-UJvUtyH:brqzjrv) ([raw file](https://github.com/vaadin/vaadin-grid-flow/blob/ca072139ef76cabfd3107697955b599e9e246989/src/main/java/com/vaadin/flow/component/grid/AbstractColumn.java#L93)):*
> ```Java
>                     .beforeClientResponse(this, context -> renderHeader()));
>         } else {
>             renderHeader();
> ```

What happens if you set a ComponentRenderer and after it set a TemplateRenderer? Looks like the rendering will be applied twice.

Suggestion: add a test with this scenario, and if it is the case that the rendering is applied twice, just remove this if/else and apply `beforeClientResponse` for both cases.

---


*Comments from [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid-flow/176)*

pekam:

---

*[src/main/java/com/vaadin/flow/component/grid/AbstractColumn.java, line 93 at r1](https://reviewable.io/reviews/vaadin/vaadin-grid-flow/176#-LBZ_oqDCllm-UJvUtyG:-LBZgQfv4nAo1DCVYZ5e:b-8sipzf) ([raw file](https://github.com/vaadin/vaadin-grid-flow/blob/ca072139ef76cabfd3107697955b599e9e246989/src/main/java/com/vaadin/flow/component/grid/AbstractColumn.java#L93)):*

What happens if you set a ComponentRenderer and after it set a TemplateRenderer? Looks like the rendering will be applied twice.

Suggestion: add a test with this scenario, and if it is the case that the rendering is applied twice, just remove this if/else and apply `beforeClientResponse` for both cases.
</blockquote></details>

Actually the only reason why I added this `instanceof` check here is the unit tests. They broke when I postponed the rendering to `beforeClientResponse`.

Yes, the rendering will be applied twice, but the correct rendering will still be applied last so it's just a bit of extra computation. Before this change the rendering was applied even more often unnecessarily.

So I don't know... I would like to remove the `if instanceof` but I don't know how to fix all my unit tests which expect that the template is in place immediately when setting the header/footer.

---


*Comments from [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid-flow/176)*

gilberto-torrezan:
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[src/main/java/com/vaadin/flow/component/grid/AbstractColumn.java, line 93 at r1](https://reviewable.io/reviews/vaadin/vaadin-grid-flow/176#-LBZ_oqDCllm-UJvUtyG:-LBZxEEZ8mYn61puccRH:b-1r1y3r) ([raw file](https://github.com/vaadin/vaadin-grid-flow/blob/ca072139ef76cabfd3107697955b599e9e246989/src/main/java/com/vaadin/flow/component/grid/AbstractColumn.java#L93)):*
<details><summary><i>Previously, pekam (Pekka Maanpää) wrote…</i></summary><blockquote>

Actually the only reason why I added this `instanceof` check here is the unit tests. They broke when I postponed the rendering to `beforeClientResponse`.

Yes, the rendering will be applied twice, but the correct rendering will still be applied last so it's just a bit of extra computation. Before this change the rendering was applied even more often unnecessarily.

So I don't know... I would like to remove the `if instanceof` but I don't know how to fix all my unit tests which expect that the template is in place immediately when setting the header/footer.
</blockquote></details>

You can probably mock the `runWhenAttached` call to run it immediately when called by the test. We have something like that in the `test-root-context` project, and Denis knows more about it.

But what about a test swapping ComponentRenderer and TemplateRenderer? Do we have one test like it already?

---


*Comments from [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid-flow/176)*

pekam:
Review status: 1 of 6 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/main/java/com/vaadin/flow/component/grid/AbstractColumn.java, line 93 at r1](https://reviewable.io/reviews/vaadin/vaadin-grid-flow/176#-LBZ_oqDCllm-UJvUtyG:-LB_Qs2CBSi9XupnnBLt:btp2tkx) ([raw file](https://github.com/vaadin/vaadin-grid-flow/blob/ca072139ef76cabfd3107697955b599e9e246989/src/main/java/com/vaadin/flow/component/grid/AbstractColumn.java#L93)):*

You can probably mock the `runWhenAttached` call to run it immediately when called by the test. We have something like that in the `test-root-context` project, and Denis knows more about it.

But what about a test swapping ComponentRenderer and TemplateRenderer? Do we have one test like it already?
</blockquote></details>

I replaced those tests with ITs and added a test for changing from text header to component header

---


*Comments from [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid-flow/176)*

gilberto-torrezan:
:lgtm:

---

Reviewed 5 of 5 files at r2.

---



*Comments from [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid-flow/176#-:-LB_Z5y99ZlQZ-XyNKhP:bnfp4nl)*

PR #202, opened 2018-05-21
Fix removing empty column-groups after initial rendering

pekam:
The logic for checking empty column-groups was to check that
childCount==0, but that worked only before first rendering (before
templates were attached beforeClientResponse). Now it checks if there
are no column or column-group children.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid-flow/202)


caalador:
Reviewed 4 of 4 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid-flow/202)*

PR #249, opened 2018-06-29
Observe changes in _previousSorters instead of using an eventListener.

### LINKED ISSUE #188: Reset sort order generates server exception
I have a grid that gets its rows from a `AbstractBackEndDataProvider`. The server throws an exception the third time that I click the header (to sort it, the third time resets the sorting). The exception is in the following code (in Grid.java, version 1.0.0-beta7): 

```java
switch (sorter.getString("direction")) {  // <-- Exception here
case "asc":
    sortOrderBuilder.thenAsc(column);
    break;
case "desc":
    sortOrderBuilder.thenDesc(column);
    break;
default:
    throw new IllegalArgumentException(
            "Received a sorters changed call from the client containing an invalid sorting direction");
}
```

`sorter.getString("direction")` return a `JreJsonNull` and that throws the exception `java.lang.ClassCastException: elemental.json.impl.JreJsonNull cannot be cast to elemental.json.JsonString`.

The third time the client sends `sorter.direction = null` but that is not allowed and it should not happen. If I run the demo code for `vaadin-grid-flow` it works fine (without the exception) but it doesn't use the `AbstractBackEndDataProvider`. I'm not sure if `AbstractBackEndDataProvider` is the problem.

I have debugged the client side code and I think the problem is in the order that two functions are called. When I run the demo code for `vaadin-grid-flow` the functions are called in the following order.

1. `vaadin-grid-combined-mixin.html  ->  _onSorterChanged(e)`
2. `gridConnector.html  ->  const sorterChangeListener = function (event)`

`_onSorterChanged(e)` will clean the variable `_sorters`. When `sortOrderChangedListener` is called there will be no `direction = null` sent to server.

When I debug my code with `AbstractBackEndDataProvider` the functions are called in reverse order (sometimes, bu not always):

1. `gridConnector.html  ->  const sorterChangeListener = function (event)`
2. `vaadin-grid-combined-mixin.html  ->  _onSorterChanged(e)`

`sortOrderChangedListener` will call server with `direction = null` and the server will throw an exception.

I don't know why the order is different and why it sometimes work and sometimes not. I guess it is something with the order that Flow send its data to the client, but that is something that I haven't managed to debug.

caalador:
This way we update the serverside sorting always after the
sorter-changed event has been handled by grid.

Fixes #188

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid-flow/249)

### DIFF
--- a/src/main/resources/META-INF/resources/frontend/gridConnector.js
+++ b/src/main/resources/META-INF/resources/frontend/gridConnector.js
@@ -371,7 +371,7 @@ window.Vaadin.Flow.gridConnector = {
       }
     }
 
-    const sorterChangeListener = function(event) {
+    const sorterChangeListener = function() {
       grid.$server.sortersChanged(grid._sorters.map(function(sorter) {
         return {
           path: sorter.path,
@@ -379,7 +379,7 @@ window.Vaadin.Flow.gridConnector = {
         };
       }));
     }
-    grid.addEventListener('sorter-changed', sorterChangeListener);
+    grid._createPropertyObserver("_previousSorters", sorterChangeListener);
 
     grid._updateItem = function(row, item) {
       Vaadin.GridElement.prototype._updateItem.call(grid, row, item);


denis-anisimov:
There is a test failure. Not sure whether it relates to the PR though.
I will rerun tests.

---

Review status: 0 of 1 files reviewed, all discussions resolved

---



*Comments from [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid-flow/249#-:-LGOL0YK0sG1mfKx2mC6:bjwczx3)*

denis-anisimov:
:lgtm:

---

Reviewed 1 of 1 files at r1.
Review status: :shipit: complete! all files reviewed, all discussions resolved

---



*Comments from [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid-flow/249#-:-LGONq031gubhlTVvVHf:bnfp4nl)*

PR #285, opened 2018-07-26
Make automatic columns sortable by default if the property is Comparable

gilberto-torrezan:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid-flow/285)


pekam:
According to the ticket this should be backported to V10, so please cherrypick this to the other branches too.

PR #308, opened 2018-08-08
Depend on all WebJars explicitly to avoid version ranges in resulting jars (#256)

denis-anisimov:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid-flow/308)


ZheSun88:
the milestone should be `1.1.1`? or @denis-anisimov

Issue #368, opened 2018-10-11
It's not possible to update rendered component if editor is used


denis-anisimov:
`Grid.Column` has one extra method `setEditorComponent`  which may be shouldn't be there at all (as a public AIP).
Initially it was just a private method.
But I made it public for reasons described below:

`GridView`  contains `createBufferedEditor` method.
To be able to show and control buffered editor there is a column with "Edit" button.

The proper implementation would be:
- has a `Div`  as a container
- add **ALL** button right away into the container (`edit`, `save`, `cancel`).
- Use `setVisible` for buttons that are shouldn't be shown: `save` and `cancel` initially and when `edit` is clicked it should become hidden and `save`/`cancel` should be visible.

Using this approach you don't need any `setEditorComponent` method.

The problem is : it doesn't work in this way. `Edit` button is always shown regardless of its visibility after row becomes rendered for editing.

Everything works fine if there is no any editor involvement. But once the editor is shown via template rendering everything explodes.

Needs to be investigated since even if we have a workaround for this it's not good.

pleku:
So to recap, for some reason, when there is a an editor component in any of the the columns, the component renderer is broken.

PR #375, opened 2018-10-18
Refactor events on Grid editor, and add close listener

gilberto-torrezan:
This patch refactors the event system on the Grid editor, making it
simpler. Also adds the close listener, and tests using the new
ComponentRenderer edit function.

Related to #368

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid-flow/375)


Legioth:
*[a discussion](https://reviewable.io/reviews/vaadin/vaadin-grid-flow/375#-LP4zsa9CAQjxLGVN6aC:-LP5qk8r6FLIwTDZ6O40:b-bujspm) (no related file):*

Let's say overall: I'm fine with this approach as long as we don't care about backward compatibility. Could you please ask Leif about it?
</blockquote></details>

As a general principle, I think it's clearer to have a distinct event type for each type of listener. This also gives us flexibility to later on introduce additional functionality to one of the events without having to do a breaking change by introducing a separate event type at that point.

It might still make sense to have a common base type for each of the different events, as is done with e.g. `com.vaadin.flow.router.BeforeEvent` and `com.vaadin.flow.component.internal.KeyboardEvent`.

Compatibility with V8 is also an added bonus :)

---

Issue #380, opened 2018-10-23
Test GridViewIT.basicRenderers_rowsAreRenderedAsExpected fails (due to locale settings?)


stefanuebe:
I forked the repo, checked out the master, created a branch and run clean install with profile "runTests". Tests failed at the following point: [ERROR]   GridViewIT.basicRenderers_rowsAreRenderedAsExpected:666->assertCellContent:1000 Wrong content of the rendered cell expected:<1[/10/18 11:19:11 AM]> but was:<1[0.01.18 11:19:11]>

Since my machine runs with locale settings for Germany, I guess this might be the reason for the "wrong" rendering results.

stefanuebe:
Solution proposal: either the test assert should use the same DateTimeFormatter as is passed to the renderers starting at lines 991 of class `com.vaadin.flow.component.grid.demo.GridView` or DTF instance for formatter should get passed a fixed english Locale but not default.

stefanuebe:
Adding `.withLocale ( Locale.ENGLISH )` to the two formatter instances made the test pass successfully.

stefanuebe:
@see https://github.com/vaadin/vaadin-grid-flow/pull/382

PR #382, opened 2018-10-23
Fix for https://github.com/vaadin/vaadin-grid-flow/issues/380. 

stefanuebe:
DateTimeFormatter instances in (test) GridView now always use english locale as is used for assertion in GridViewIT.


stefanuebe:
Test update not needed, can be fixed with passing modified locale settings to VM.

PR #417, opened 2018-11-08
Unregister column data generators if column is invisible

### LINKED ISSUE #117: Grid sends data related to hidden columns, when it shouldn't
Currently when a column is set to hidden (`setVisible(false)`), the data needed to render its items is always sent to the client - when it actually should never send them.

This creates some unnecessary impact on performance (mainly on old browsers) when a Grid has several hidden columns.

denis-anisimov:
Fixes #117

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid-flow/417)


denis-anisimov:
Closed in favor of #424

PR #453, opened 2018-12-07
Add Grid static web resources OSGi contributor (#449)

### LINKED ISSUE #440: OSGi: gridConnector.js and vaadin-grid-flow-selection-column.html need to be registered as a web resource
`gridConnector.js` and  `vaadin-grid-flow-selection-column.html`  are not registered as  web resources.
As a result it's not available at runtime via HTTP.
During webjar unpacking it's easier to forget about it and not unpack.
As a result the functionality which depends on it won't work.

It can be unpacked : yes ( same way as webjars are unpacked).
But there is no reason to do this. We can just register them as a `OsgiVaadinContributor` (similar way as flow-client registers resources).

denis-anisimov:
Fixes #440

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid-flow/453)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -107,6 +107,13 @@
         </dependency>
          <dependency>
             <groupId>com.vaadin</groupId>
+            <artifactId>flow-osgi</artifactId>
+            <version>${flow.version}</version>
+            <scope>provided</scope>
+        </dependency>
+        
+        <dependency>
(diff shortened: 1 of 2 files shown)


denis-anisimov:
This should not be in `1.2` since it doesn't have required OSGi support.

PR #474, opened 2018-12-21
Add SubMenu API for GridContextMenu

pekam:
Depends on https://github.com/vaadin/vaadin-context-menu-flow/pull/36

Breaking changes:
- `GridContextMenu` is moved to package `com.vaadin.flow.component.contextmenu`
- `GridContextMenu::addItem` returns an object of type `GridMenuItem<>` instead of `MenuItem`
- Use the latest `vaadin-context-menu-flow` which introduces breaking changes listed in https://github.com/vaadin/vaadin-context-menu-flow/pull/36

Because of the breaking changes, the version is bumped to 3.0

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid-flow/474)


pleku:
Are the breaking changes mandatory? Are you sure? Can you keep from breaking things and deprecate the things that should be changed?

pekam:
The class is moved to contextmenu package because it uses a package-protected helper class from there. This can be avoided by duplicating that helper class in Grid repository. Then we'd also need to duplicate a bunch of tests in two repos. Also, now that there are many contextmenu-related classes in Grid repository, it's cleaner to have them in a separate package. So the first breaking change can be avoided at the cost of making the maintenance harder.

The second breaking change could be avoided by making MenuItem a superclass of GridMenuItem. Then the common menu item would need to extend that also, so ContextMenu::addItem would need to return some other type, like BasicMenuItem. This would be a compromise in the API and we concluded with @Legioth that it's better to break things at this point.

pleku:
>GridContextMenu is moved to package com.vaadin.flow.component.contextmenu

>The class is moved to contextmenu package because it uses a package-protected helper class from there. This can be avoided by duplicating that helper class in Grid repository. Then we'd also need to duplicate a bunch of tests in two repos. Also, now that there are many contextmenu-related classes in Grid repository, it's cleaner to have them in a separate package. So the first breaking change can be avoided at the cost of making the maintenance harder.

This will BTW break OSGi compatibility; you cannot have two artifacts that export classes from the same package.

PR #481, opened 2019-01-18
No header row should be added if not defining column header(s).

### LINKED ISSUE #477: V12+ Grid gets an empty header row
If no setHeader is specified for any Grid column then the expectation would be that no header row is drawn.

With the current version the header row is drawn as empty string.

ColumnPathRenderer sets the header property to "" which made the header column not be available, but after https://github.com/vaadin/vaadin-grid/issues/1454 this makes us get an empty header string.

This should be updated to use `container.setPropertyJson("header", Json.createNull());` after vaadin-grid updates with the feature that removes default header when "header = null" [Simple column does not allow invisible header](https://github.com/vaadin/vaadin-grid/issues/1540)

caalador:
Fixes #477

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid-flow/481)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -33,6 +33,7 @@
         <dependency>
             <groupId>org.webjars.bowergithub.vaadin</groupId>
             <artifactId>vaadin-grid</artifactId>
+            <version>5.2.7</version>
         </dependency>
 
         <dependency>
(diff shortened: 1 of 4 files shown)


caalador:
Closing as faulty.
Master requires new alpha

PR #484, opened 2019-01-21
Add overloads for addColumns

yuriy-fix:
Required by https://github.com/vaadin/vaadin-grid-pro-flow/pull/7

### DIFF
--- a/src/main/java/com/vaadin/flow/component/grid/Grid.java
+++ b/src/main/java/com/vaadin/flow/component/grid/Grid.java
@@ -28,6 +28,7 @@ import java.util.NoSuchElementException;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
+import java.util.function.BiFunction;
 import java.util.function.BinaryOperator;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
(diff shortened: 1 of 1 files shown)


Legioth:
*[a discussion](https://reviewable.io/reviews/vaadin/vaadin-grid-flow/484#-LWoMg-l-4_puiX285vD:-LWpQm1m4O4n5MZzGFbv:bqrsz07) (no related file):*

>  but then also use the existing this::createColumn as the default factory implementation that would be used in cases where no specific factory is provided.

Definitely.

>I don't think it's necessary to deprecate createColumn since it helps simplify situations where the same customization is always needed.

`createColumn`  is very confusing. Since it's in the `Grid` class the contract implicitly suggests that it will be used somehow. But it's not used at all anyhow when a custom column factory is passed.

I agree that the default column factory should be available in cases when this default factory needed in similar situations.
But there are many ways to do it.

First of all : I would avoid having it as a method which already is a factory. Instead I would like to have something which returns a factory.

So there should be either:
- a `static final protected` constant `DEFAULT_FACTORY` whose value is a factory (a lambda e.g.)
- a method which returns a factory (the same thing but using a method contract instead of a field) . And its name should be `getDefaultSomething`.

It allows to clarify the contract so it doesn't introduce a confusion. Now it's just a (default) factory which doesn't have to be used (be may be used).
</blockquote></details>

Good points.  Based on that, I'd suggest something like `getDefaultColumnFactory()` as a protected method that is intended to be overridden. We could then deprecate `createColumn` but for the time being implement `getDefaultColumnFactory()` to still use `createColumn` so that any existing Grid subclass implementation that overrides `createColumn` would keep working.

---

PR #508, opened 2019-02-06
fix connector so as it works in Polymer 3

### LINKED ISSUE #509: vaadinConnector does not work in P3
In Polymer3 `Vaadin.Grid` namespace does not work anymore, because the way to create a new grid or any other element provided by the grid should be by using ES6 imports e.g. `import  { Grid } from '@vaadin/vaadin-grid/vaadin-grid.js'`

`vaadinConnector` is monkey-patching the `ItemCache` via the `Vaadin.Grid.ItemCache.prototype` and makes P3 apps fail.  Thus, the ideal solution should be `import { ItemCache } from '@vaadin/vaadin-grid/vaadin-grid-data-provider-mixin.js'`

Another problem in gridConnector is that it's patching the prototype each time a new grid is instantiated, it might be by mistake, but anyways it's always better to monkey-patch the instance instead of the prototype, otherwise if the user mixes pure vaadin-grid in client side might break it.

manolo:
Fixes #509 

This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid-flow/508)

### DIFF
--- a/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
+++ b/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
@@ -4,8 +4,36 @@ window.Vaadin.Flow.gridConnector = {
     if (grid.$connector){
       return;
     }
+    grid.$connector = {};
 
-    Vaadin.Grid.ItemCache.prototype.ensureSubCacheForScaledIndex = function(scaledIndex) {
+    function createCache(parentCache, parentItem) {
+      // TODO(manolo) ugly, we need to change it when we can use ES6 imports
+      // as indicated in https://github.com/vaadin/vaadin-grid-flow/issues/509
+      const cache =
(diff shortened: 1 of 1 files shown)


vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
```diff
@@ -5,7 +5,7 @@ window.Vaadin.Flow.gridConnector = {
       return;
     }
 
-    Vaadin.Grid.ItemCache.prototype.ensureSubCacheForScaledIndex = function(scaledIndex) {
+    grid._cache.ensureSubCacheForScaledIndex = function(scaledIndex) {
```
tomivirkki: I think this will only patch the first level cache. But with hiearachical grid, all the levels have their own cache instances and those remain unpatched.

manolo:
closing, in favor of #510 which has a less intrusive implementation

Issue #515, opened 2019-02-10
Uncaught TypeError on inital load


appreciated:
A user of an addon of mine (App-Layout) [reported an issue](https://github.com/appreciated/vaadin-app-layout/issues/213) which seems to be connected with an unexpected state in the [`grid-connector.js`](https://github.com/vaadin/vaadin-grid-flow/blob/master/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js#L544-L561).

[A minimal example can be found here](https://github.com/jflamy/bug_crud.git)
- Run `org.ledocte.owlcms.Main`
- Navigate to `http://127.0.0.1:8080/main/test`

Browser console Stacktrace:
```
client-28831013082792340517B7074033EF04.cache.js:54 Uncaught TypeError: Cannot read property 'map' of null
    at updateGridItemsInDomBasedOnCache (gridConnector.js:556)
    at Object.grid.$connector.set (gridConnector.js:590)
    at Object.eval (eval at jt (client-28831013082792340517B7074033EF04.cache.js:975), <anonymous>:3:15)
    at jt (client-28831013082792340517B7074033EF04.cache.js:975)
    at it (client-28831013082792340517B7074033EF04.cache.js:931)
    at gt (client-28831013082792340517B7074033EF04.cache.js:562)
    at Pq (client-28831013082792340517B7074033EF04.cache.js:482)
    at lr.mr [as W] (client-28831013082792340517B7074033EF04.cache.js:983)
    at eA (client-28831013082792340517B7074033EF04.cache.js:866)
    at pv (client-28831013082792340517B7074033EF04.cache.js:976)
    at Zw.$w [as J] (client-28831013082792340517B7074033EF04.cache.js:983)
    at Function.yl (client-28831013082792340517B7074033EF04.cache.js:983)
    at d (client-28831013082792340517B7074033EF04.cache.js:640)
    at Set.forEach (<anonymous>)
    at bl (client-28831013082792340517B7074033EF04.cache.js:740)
    at HTMLElement.s.ready (client-28831013082792340517B7074033EF04.cache.js:983)
    at HTMLElement._enableProperties (properties-changed.html:321)
    at HTMLElement.connectedCallback (properties-mixin.html:219)
    at HTMLElement.connectedCallback (element-mixin.html:595)
    at nv (client-28831013082792340517B7074033EF04.cache.js:941)
    at Fv (client-28831013082792340517B7074033EF04.cache.js:902)
[...]

PR #517, opened 2019-02-11
Fixes #516 and #515

### LINKED ISSUE #516: gridConnector.js (cannot read property 'map' of null)
When using two very useful extensions together (crudui and app-layout) a Grid bug becomes visible -- a message (cannot read property 'map' of null) is reported.
See [bug report and minimal reproducible setup](https://github.com/alejandro-du/crudui/issues/45).

After investigation "appreciated", the author of app-layout, has identified a fix, see the [following screenshot](https://user-images.githubusercontent.com/1729845/52534538-129f4f80-2d43-11e9-833a-0fba3ce61840.PNG) for his suggested modifications.

appreciated:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid-flow/517)

### DIFF
--- a/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
+++ b/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
@@ -542,7 +542,7 @@ window.Vaadin.Flow.gridConnector = {
      * @param array items the items to update in DOM
      */
     const updateGridItemsInDomBasedOnCache = function(items) {
-      if (!items) {
+      if (!items || !grid._physicalItems) {
         return;
       }
       /**


appreciated:
> I would have done the same changes myself but didn't yet get to it.
@pleku I'm sure you would have 🙂. 
> Also I had no idea (nor did @tomivirkki ) that this might get executed when the underlying iron list doens't have anything in the DOM.
Maybe the <app-layout> causes some additional resize events which then cause the grid to update itself before it is initialized 🤔? Just a guess.

appreciated:
@denis-anisimov not sure since this only throws a client side exception. Note that I have pretty much no experience with Testbench.

heruan:
@denis-anisimov There is an example project in #515 which should replicate the issue, is that what you need?

denis-anisimov:
> @denis-anisimov There is an example project in #515 which should replicate the issue, is that what you need?

OK, so it means that it's possible to make IT test for the fix.

heruan:
@appreciated and @jflamy do you think you can add an IT to get this merged? You already have a sample project which replicates, I have no TestBench experience but I guess it should be pretty straighforward with the code you already have.

heruan:
Worth noting this could be fixed in the Web Component, initializing `_physicalItems` as an empty array instead of `null` at [vaadin/vaadin-grid/src/iron-list.html#L113](https://github.com/vaadin/vaadin-grid/blob/63d3f80232a4737d4ae234795c6b076dcd72d270/src/iron-list.html#L113).

appreciated:
Any requirements for the integration-test? Can I find some coding guidelines for that anywhere?

denis-anisimov:
No any specific requirements.
Let's see the result in review.

Just use existing code as an example.
E.g. https://github.com/vaadin/vaadin-grid-flow/blob/master/vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/it/DetailsGridPage.java is a page which contains a usecase (UI control elements to reproduce the issue).
And https://github.com/vaadin/vaadin-grid-flow/blob/master/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/DetailsGridIT.java is an IT test code.

heruan:
@appreciated, did you find out what is triggering the error? I'm not using any addon (just `Grid`) and I'm experiencing this in my app, but I cannot replicate in a simple project or IT test.

heruan:
I still can't figure out what is causing this. Any chance the team will fix this for V13, if the author doesn't?

appreciated:
I'm onto something. This issue only if the grid is being wrapped inside a Polymer Template.

appreciated:
This is so wierd 😒. 

wrapping the Grid in a Template as follows is enough.
### MyTemplate 
```
@Tag("a-template")
@HtmlImport("frontend://src/polymer-template.html")
public class MyTemplate extends PolymerTemplate<TemplateModel> implements HasComponents {
    @Id("test")
    Div div;
}
```

### polymer-template.html
```
<link rel="import" href="../bower_components/polymer/polymer-element.html">

<dom-module id="a-template">
    <template>
        <div id="test">
        <slot></slot>
        </div>
    </template>
    <script>
        class ATemplate extends Polymer.Element {
            static get is() {
                return 'a-template'
            }
        }
        customElements.define(ATemplate.is, ATemplate);
    </script>
</dom-module>
```

appreciated:
@denis-anisimov I just found that I'm required to have a pro subscription to write Integration tests which I currently don't have. So I guess this need to be done by someone that has one 😓.

heruan:
Thank you @appreciated for the thorough analysis! In my case the grid is in the lightdom of a component which is a `PolymerTemplate`, so maybe you're right.

I really hope the team can merge this for the upcoming platform release!

heruan:
Worth noting that another issue in `ComboBox` connector happens under similar conditions: https://github.com/vaadin/vaadin-combo-box-flow/issues/219

denis-anisimov:
I cannot merge it without a test.
It's too hard to use your branch to write the test and the fix is trivial. 
So it's better to close this PR and make a new one.

denis-anisimov:
It's decided to merge this PR.
I will create another ticket for a test.

PR #623, opened 2019-04-16
feat: column auto-width

### LINKED ISSUE #619: API for setting autoWidth on column
Depends on https://github.com/vaadin/vaadin-grid/issues/1233

Haprog:
New feature for automatically setting the width of a column based on the column contents. For more info see the new demo "Configuring columns / Columns with automatic width" and the Javadoc of `Column#setAutoWidth(boolean)`.

Fixes #619

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid-flow/623)

### DIFF
--- a/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
+++ b/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
@@ -415,6 +415,50 @@ public class Grid<T> extends Component implements HasDataProvider<T>, HasStyle,
         }
 
         /**
+         * Enables or disables automatic width for this column.
+         * <p>
+         * Automatically sets the width of the column based on the column
+         * contents when this is set to {@code true}.
+         * <p>
+         * For performance reasons the column width is calculated automatically
+         * only once when the grid items are rendered for the first time and
+         * the calculation only considers the rows which are currently rendered
+         * in DOM (a bit more than what is currently visible). If the grid is
+         * scrolled, or the cell content changes, the column width might not
+         * match the contents anymore.
+         * <p>
+         * Hidden columns are ignored in the calculation and their widths are
+         * not automatically updated when you show a column that was initially
+         * hidden.
+         * <p>
+         * You can manually trigger the auto sizing behavior again by calling
+         * {@see Grid#recalculateColumnWidths()}.
+         * <p>
+         * The column width may still grow larger when {@code flexGrow} is not
+         * 0.
+         *
+         * @see Grid#recalculateColumnWidths()
+         *
+         * @param autoWidth
+         *            whether to enable or disable automatic width on this column
+         * @return this column, for method chaining
+         */
+        public Column<T> setAutoWidth(boolean autoWidth) {
+            getElement().setProperty("autoWidth", autoWidth);
+            return this;
+        }
+
+        /**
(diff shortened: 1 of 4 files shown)


vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
```diff
@@ -409,6 +409,50 @@ public int getFlexGrow() {
...
+         * in DOM (a bit more than what is currently visible). If the grid is
+         * scrolled, or the cell content changes, the column width might not
+         * match the contents anymore.
+         * <p>
+         * Hidden columns are ignored in the calculation and their widths are
+         * not automatically updated when you show a column that was initially
+         * hidden.
+         * <p>
+         * You can manually trigger the auto sizing behavior again by calling
+         * {@see Grid#recalculateColumnWidths()}.
```
pekam: The `@see` doesn't work in the middle of the description. It should be used only as a separate mention and it renders a "See also: "-message before the link.

Fix this by changing `see` to `link`.

vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
```diff
@@ -409,6 +409,50 @@ public int getFlexGrow() {
...
+         * <p>
+         * The column width may still grow larger when {@code flexGrow} is not
+         * 0.
+         *
+         * @see Grid#recalculateColumnWidths()
+         *
```
pekam: `@see Column#setFlexGrow` could be added here as `flexGrow` is mentioned above.

vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
```diff
@@ -409,6 +409,50 @@ public int getFlexGrow() {
...
+            getElement().setProperty("autoWidth", autoWidth);
+            return this;
+        }
+
+        /**
+         * Gets this column's auto width state.
+         *
+         * @return whether this column has automatic width enabled
+         */
+        @Synchronize("auto-width-changed")
```
pekam: This annotation defines that the property value should be synchronized to the server when a `auto-width-changed`-property has changed. Based on the client-side implementation it looks like there's no `auto-width-changed` event, so this line is redundant. Also, we can expect a Java user to only set this value via the Java-API, so we don't have to worry this property getting out-of-sync by something changing it in the client-side.

vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
Haprog: Done.

vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
Haprog: Done.

vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
Haprog: Removed this annotation, it didn't seem to work anyway.

PR #643, opened 2019-05-10
Remove checkbox line breaks from selection column

tomivirkki:
Fixes the issue with unexpected ellipsis

Screenshot 2019-05-10 14 21 48

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid-flow/643)


tomivirkki:
I think the issue was more about grid flow installing an outdated package of vaadin-checkbox. This issue is fixed in the newest checkbox and shouldn't happen anymore as long as a proper version is installed. Closing.

PR #669, opened 2019-06-07
Fix grid build

caalador:
make legacy test more secure

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid-flow/669)


caalador:
Build fixed in #674

PR #700, opened 2019-07-01
Reset requested range on ArrayUpdater.init

### LINKED ISSUE #539: Data missing if detached and attached when scrolled down
If a Grid instance is scrolled down, detached and then attached back again, then there there won't be any data in the grid. The same doesn't happen if only scrolling down by a smaller amount.

```
@Route
public class MainView extends VerticalLayout {
    public MainView() {
        Grid<String> grid = new Grid<>();

        grid.addColumn(string -> string).setHeader("Value");
        grid.setItems(IntStream.range(0, 150).mapToObj(Integer::toString));

        Button toggleButton = new Button("Toggle visibility", event -> {
            if (grid.getParent().isPresent()) {
                remove(grid);
            } else {
                add(grid);
            }
        });

        add(new Text("Scroll grid to the bottom. " 
                + "Click the toggle button twice. "
                + "Expected that grid still has contents, but it does not."),
            toggleButton, grid);
    }
}
```

Tested with Vaadin 12.0.6 and 13.0.0.beta1.

pekam:
Fix #539

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid-flow/700)

### DIFF
--- a/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
+++ b/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
@@ -1073,6 +1073,7 @@ public class Grid<T> extends Component implements HasDataProvider<T>, HasStyle,
         public void initialize() {
             initConnector();
             updateSelectionModeOnClient();
+            setRequestedRange(0, getPageSize());
         }
 
         @Override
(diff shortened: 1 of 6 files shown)


vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
```diff
@@ -1073,6 +1073,7 @@ public UpdateQueue startUpdate(int sizeChange) {
         public void initialize() {
             initConnector();
             updateSelectionModeOnClient();
+            setRequestedRange(0, getPageSize());
```
yuriy-fix: We should consider the same for `TreeGrid`.

pekam: Good catch! Done.

PR #717, opened 2019-07-21
DetailsRenderer and ClassNameGenerator in the same Grid fails

whiteancient:
If I have a column with a ClassNameGenerator in a Grid where I also have a DetailsRenderer,
The code accesses the ClassNameGenerator-method with an undefined column. This is a solution to that problem.
It might be a bandaid on another problem, but it at least solves it.
(This is my first pull-request ever, tell me if I do something wrong)

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid-flow/717)

### DIFF
--- a/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
+++ b/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
@@ -922,10 +922,10 @@ window.Vaadin.Flow.gridConnector = {
 
     grid.cellClassNameGenerator = function(column, rowData) {
         const style = rowData.item.style;
-        if (!style ||!column) {
+        if (!style) {
             return;
         }
-        return (style.row || '') + ' ' + (style[column._flowId] || '');
+        return (style.row || '') + ' ' + ((column && style[column._flowId]) || '');
     }
 
     grid.dropFilter = rowData => !rowData.item.dropDisabled;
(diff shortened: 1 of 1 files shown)


pekam:
Thanks for the PR!

It seems that the issue is a regression caused by #695, by not considering that the client-side class name generator will be run also for details rows with an `undefined` column.

Before #695, Grid would set the class names of the row also to the details row. Your PR would change it so that details rows don't get the generated class names.

There are two options for handling this, depending on whether the generator should be applied for details rows as well.

1. If yes: the PR should be changed so that it doesn't return empty if column is not defined, but instead just skips trying set the column class, so that the row class is still applied, `((column && style[column._flowId]) || '')`, so it works like previously.
2. If no: the fix should be done [in the web component repository](https://github.com/vaadin/vaadin-grid/blob/f6efd4ea2c331a40451d25667527cf7488e2403e/src/vaadin-grid-styling-mixin.html#L59), so that it skips generating classes for details rows.

Originally it was planned, that [details rows should not be affected by the generator](https://github.com/vaadin/vaadin-grid/issues/1297#issuecomment-428841155), but this has been ignored in the actual implementation, and using the class name generator with details rows seems to be not tested nor documented.

I think we have to go with option 1, because the functionality can be easily thought as a feature instead of a bug by existing users. We might have web component users who generate classes for the details rows as well by checking the `column` argument. Existing Flow users also may expect that the classes are generated for the details and already use that feature. Flow users would also still be able to disable class names from details rows with CSS selectors. If we don't change the client-side functionality, we should still document that `column` may be `undefined` in case of details rows, and add tests for this.

whiteancient:
I believe I have now updated the PR to your recommendation.

PR #729, opened 2019-07-30
Fix a problem in running ITs locally

mehdi-vaadin:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid-flow/729)


alvarezguille:
This shouldn't be needed unless we have incorrect RunOnHub annotation in tests

PR #731, opened 2019-08-05
Setting details row visible doesn't work initially 

jcgueriaud1:
Initial behaviour:
* Details rows are deleted when the data of the grid is refreshed.

Changes:
* Details row are kept if deleteAll is called.

Add test cases for the grid and the treegrid.

Requires this change ( https://github.com/vaadin/flow/pull/6144 ) to be merged in flow or the details will displayed but empty.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid-flow/731)


jcgueriaud1:
The fix is in #732 
Closed as duplicated

Issue #701, opened 2019-07-02
TreeGrid with details is unreliable in Chrome, a little less in Safari, less but still unreliable in Firefox


enver-haase:
This relates to #563, #525 .

I have a hard time describing what the TreeGrid actually does but it does not do things right when it comes to details rows.

It seems to work fine in Firefox until I collapse the root for the first time.

In Safari, I see mostly empty details rows -- but not only.

Can you please check out the code at
https://github.com/enver-haase/Playground/tree/master/summarizing-tree
(very simple TreeGrid) and see yourself?

I am happy to provide an online demonstration.

enver-haase:
Oh, used Vaadin 14.0.0.rc3

jcgueriaud1:
Hi,

In your example, there is 2 bugs:
* the #525 bug (the item detail is not rendered correctly the first time)
* when you collapse a node then:
** item details is not shown (but everything is processed correctly)

But thanks for your code, it's easy to reproduce the wrong behaviour.

jcgueriaud1:
Closed by #732

Issue #525, opened 2019-02-14
setting details row visible doesn't work initially


OlliTietavainenVaadin:
`<vaadin.version>12.0.5</vaadin.version>`
If you call `grid.setDetailsVisible(item, true)` when you're initializing the Grid, it doesn't open the details row(s). Calling the method later (for example, from a button's click listener) opens the details rows as expected.

Sample code that demonstrates the issue:
```java
    Grid<Bean> grid = new Grid<>();
    List<Bean> items = new ArrayList<>();

    public void setFirstAndSecondItemsVisible() {
        grid.setDetailsVisible(items.get(0), true);
        grid.setDetailsVisible(items.get(1), true);
    }

    public MainView() {
        for (int i = 0; i < 5; i++) {
            Bean bean = new Bean(i+"");
            items.add(bean);
        }
        ListDataProvider<Bean> ldp = new ListDataProvider<>(items);
        grid.setDataProvider(ldp);
        Grid.Column<Bean> nameColumn = grid.addColumn(bean -> bean.getName()).setHeader("name");
        grid.setItemDetailsRenderer(new ComponentRenderer<Button, Bean>(item -> new Button(item.getName())));
        setFirstAndSecondItemsVisible(); // should open first and second details rows
        add(grid,
                new Button("click to open details", e -> {
                    setFirstAndSecondItemsVisible(); // same method call as above
                })
        );
    }

    public static class Bean {
        public String getName() {
            return name;
        }
        public void setName(String name) {
            this.name = name;
        }
        private String name;
        public Bean(String name) {
            this.name = name;
        }
    }
 ```

eriklumme:
As  a workaround, one can create a `Composite<Grid>` wrapper with a `@ClientCallable` method that displays the details.

    private class GridWrapper extends Composite<MyGrid> {
        @ClientCallable
        public void init() {
            itemsToDisplay.forEach(item -> getGrid().setDetailsVisible(item, true));
        }

        public EventsGrid getGrid() {
            return getContent();
        }

        @Override
        protected void onAttach(AttachEvent attachEvent) {
            super.onAttach(attachEvent);
            getContent().getElement().callFunction("$server.init");
        }
    }

To have more control over the grid creation, the `Composite` can override `initContent()`.

enver-haase:
Been horribly bitten by it today in 14.0.0.rc3

jcgueriaud1:
Closed by #732

PR #732, opened 2019-08-07
Treegrid details is not working after collapse then expand a node

jcgueriaud1:
This pull request fixes:
- #701 
and #525  (PR: https://github.com/vaadin/vaadin-grid-flow/pull/731 )

The cache was not clear correctly when a node is collapsed. So it was not rebuilt when the node expands.
There may be a better way to clear the cache.

Requires this change ( vaadin/flow#6144 ) to be merged in flow or the details will displayed but empty.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid-flow/732)

### DIFF
--- a/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
+++ b/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
@@ -218,7 +218,7 @@ window.Vaadin.Flow.gridConnector = {
       }
       // when grid is attached, newVal is not set and oldVal is undefined
       // do nothing
-      if ((newVal == null) && (typeof oldVal === 'undefined')) {
+      if ((newVal == null) && (oldVal === 'undefined')) {
         return;
       }
       if (newVal && !newVal.detailsOpened) {
--- a/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
+++ b/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
@@ -463,14 +463,9 @@ window.Vaadin.Flow.gridConnector = {
         if (parentCache && parentCache.itemkeyCaches && parentCache.itemkeyCaches[parentKey]) {
           delete parentCache.itemkeyCaches[parentKey];
         }
-        if (parentCache && parentCache.itemCaches) {
-          const keys = Object.keys(parentCache.itemCaches);
-          for (var i = 0; i < keys.length; i++) {
-            if ( parentCache.items[keys[i]].key === parentKey) {
-              delete parentCache.itemCaches[keys[i]];
-            }
-          }
-        }
+        Object.keys(parentCache.itemCaches)
+            .filter(idx => parentCache.items[idx].key === parentKey)
+            .forEach(idx => delete parentCache.itemCaches[idx]);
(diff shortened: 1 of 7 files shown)


vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
```diff
@@ -216,6 +216,11 @@ window.Vaadin.Flow.gridConnector = {
       if(!detailsVisibleOnClick) {
         return;
       }
+      // when grid is attached, newVal is not set and oldVal is undefined
+      // do nothing
+      if ((newVal == null) && (typeof oldVal === 'undefined')) {
```
tomivirkki: nit: oldVal is declared so no need to use typeof

jcgueriaud1: I removed typeof, Not sure if it's ok.

tomivirkki: Now it's comparing the value to a string 'undefined'. Should say `=== undefined`

jcgueriaud1: done

vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
```diff
@@ -458,11 +463,15 @@ window.Vaadin.Flow.gridConnector = {
         if (parentCache && parentCache.itemkeyCaches && parentCache.itemkeyCaches[parentKey]) {
           delete parentCache.itemkeyCaches[parentKey];
         }
-        if (parentCache && parentCache.itemCaches && parentCache.itemCaches[parentKey]) {
-          delete parentCache.itemCaches[parentKey];
+        if (parentCache && parentCache.itemCaches) {
+          const keys = Object.keys(parentCache.itemCaches);
```
tomivirkki: The change works well and overall lgtm. Any idea why the tests aren't passing (would have approved this otherwise)?

nit: Logic here took me a while to understand. To improve readability it could be changed to:
```
Object.keys(parentCache.itemCaches)
        .filter(idx => parentCache.items[idx].key === parentKey)
        .forEach(idx => delete parentCache.itemCaches[idx]);
```

jcgueriaud1: I changed this. I don't know why the tests are not passing.
On my computer, I've got some errors on the grid tests but not the same.
I tried treegridhuge test, and (if I removed ObjectSizeCalculator to try it) I don't have this error. I try it manually and it seems ok. I'll check the result after this modification.

tomivirkki: I think the wrapping if-statement is still needed?

jcgueriaud1: Yes, it still needed, even if no test is checking this. (never push code before morning coffee).
I increased the timeout for the test to test if it's working but slower than before.

jcgueriaud1:
This closes #701 and #751

PR #748, opened 2019-08-23
feature/dynamic context menu items

gatanaso:
This PR builds on the updates introduced in [vaadin-context-menu](https://github.com/vaadin/vaadin-context-menu-flow/pull/99), to allow for dynamically modifying the contents of the grid context menu. 

This is achieved by utilizing the newly introduced `onBeforeOpenMenu` method, where the client side event details are received and the contents of the context can potentially be modified based on that information.

This PR is the proposed solution for the following [issue](https://github.com/vaadin/vaadin-context-menu-flow/issues/76)

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid-flow/748)

### DIFF
--- a/vaadin-grid-flow-demo/src/main/java/com/vaadin/flow/component/grid/demo/GridDemo.java
+++ b/vaadin-grid-flow-demo/src/main/java/com/vaadin/flow/component/grid/demo/GridDemo.java
@@ -1750,8 +1750,6 @@ public class GridDemo extends DemoView {
     // Dynamic Context Menu begin
     private void createDynamicContextMenu() {
         TaskData taskData = new TaskData();
-        List<Task> taskList = taskData.getTasks();
-        DateTimeFormatter formatter = DateTimeFormatter.ofPattern("MM/dd/yyyy");
 
         // begin-source-example
         // source-example-heading: Dynamic Context Menu
@@ -1763,7 +1761,7 @@ public class GridDemo extends DemoView {
         grid.addColumn(Task::getName).setHeader("Task Name");
         grid.addColumn(Task::getDueDate).setHeader("Due Date");
         GridContextMenu<Task> contextMenu = new GridContextMenu<>(grid);
-        contextMenu.setContextMenuDynamicContentRenderer((task) -> {
+        contextMenu.setContextMenuDynamicContentRenderer(task -> {
             if (task == null) {
                 // do not show the context menu when a row is not clicked
                 return false;
--- a/vaadin-grid-flow-demo/src/main/java/com/vaadin/flow/component/grid/demo/GridDemo.java
+++ b/vaadin-grid-flow-demo/src/main/java/com/vaadin/flow/component/grid/demo/GridDemo.java
@@ -1,13 +1,35 @@
 package com.vaadin.flow.component.grid.demo;
 
+import java.math.BigDecimal;
+import java.text.NumberFormat;
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+import java.time.format.FormatStyle;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.Set;
+import java.util.WeakHashMap;
+import java.util.stream.Collectors;
+
+import org.apache.commons.lang3.StringUtils;
+
 import com.vaadin.flow.component.ComponentEventListener;
 import com.vaadin.flow.component.button.Button;
 import com.vaadin.flow.component.checkbox.Checkbox;
 import com.vaadin.flow.component.combobox.ComboBox;
 import com.vaadin.flow.component.datepicker.DatePicker;
 import com.vaadin.flow.component.dependency.HtmlImport;
-import com.vaadin.flow.component.formlayout.FormLayout;
 import com.vaadin.flow.component.dependency.JsModule;
+import com.vaadin.flow.component.formlayout.FormLayout;
 import com.vaadin.flow.component.grid.ColumnTextAlign;
 import com.vaadin.flow.component.grid.FooterRow;
 import com.vaadin.flow.component.grid.Grid;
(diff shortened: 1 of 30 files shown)


gatanaso:
This PR depends on the following ones being merged:
- https://github.com/vaadin/vaadin-context-menu-flow/pull/100
- https://github.com/vaadin/vaadin-context-menu-flow/pull/99
- https://github.com/vaadin/vaadin-grid-flow/pull/749

vaadin-grid-flow-demo/src/main/java/com/vaadin/flow/component/grid/demo/GridDemo.java
```diff
@@ -1740,11 +1742,41 @@ private void createContextSubMenu() {
...
+                .ofCollection(taskData.getTasks());
+
+        grid.setDataProvider(dataProvider);
+        grid.addColumn(Task::getName).setHeader("Task Name");
+        grid.addColumn(Task::getDueDate).setHeader("Due Date");
+        GridContextMenu<Task> contextMenu = new GridContextMenu<>(grid);
+        contextMenu.setContextMenuDynamicContentRenderer(task -> {
+            if (task == null) {
+                // do not show the context menu when a row is not clicked
+                return false;
```
yuriy-fix: I am not sure if that is the proper approach for handling the opening of the `contextMenu` with returning `true / false`, however I cannot propose anything better.

As discussed, probably completely ignoring the return value would be a better option.

**Not a blocker**

vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/contextmenu/DynamicContextMenuGridPage.java
```diff
@@ -0,0 +1,56 @@
+/*
+ * Copyright 2000-2018 Vaadin Ltd.
```
yuriy-fix: nit: update to `2019`

vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/contextmenu/DynamicContextMenuGridPage.java
```diff
@@ -0,0 +1,56 @@
...
+        grid.addColumn(Person::getFirstName).setHeader("Name").setId("Name-Id");
+        grid.addColumn(Person::getAge).setHeader("Born").setId("Born-Id");
+
+        grid.setItems(IntStream.range(0, 50)
+                .mapToObj(i -> new Person("Person " + i, i)));
+
+        GridContextMenu<Person> contextMenu = grid.addContextMenu();
+
+        contextMenu.setContextMenuDynamicContentRenderer(person -> {
+            if (person.getAge() < 30) {
```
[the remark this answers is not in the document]
gatanaso: Yes that is a valid point, although I was not expecting that in the test but I guess it's good to add that check.

vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/contextmenu/DynamicContextMenuGridPage.java
gatanaso: good catch 👍

vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/contextmenu/GridContextMenu.java
```diff
@@ -202,4 +207,58 @@ public Registration addGridContextMenuOpenedListener(
...
+     *
+     * The boolean return value of this callback specifies if the context menu
+     * will be opened.
+     * </p>
+     *
+     * @param contextMenuDynamicContentRenderer
+     *            the callback function that will be executed before opening the
+     *            context menu.
+     */
+    public void setContextMenuDynamicContentRenderer(
```
pekam: I have some concerns about the naming of these new methods.

First of all, the "ContextMenu"-part in the method seems redundant, as the method is in the `ContextMenu`-class, so it's already clear where the content will be rendered. Also, the terms "Dynamic" and "Content" are kind of included in the renderer-concept. So IMO the method could as well be called `setRenderer` without losing any information.

But then we come to the bigger problem: this renderer-function is not conceptually a same kind of renderer as in other Java components in Vaadin. When using renderers, the component does the actual rendering for you, while the user can use a higher-level API, eg. to provide the components or template strings that should be rendered. Here the function is just a callback, which you can use to manually remove and add components to the menu, or do any other side-effects you want.

Actually, because of the `boolean` return value, the method should technically be called `setOpenableProvider` or `setEnabledProvider`, or something like that. The rendering is still just a side-effect that the user can choose to do or not.

But if we ignore that part, the function is basically a callback to execute before the opening. So IMO it should be an event listener. But it should still be possible to prevent the menu from opening based on the item, so the event could have a method for that. This is how I'd imagine the API to be used:
```java
contextMenu.addBeforeOpenListener(event -> {
  if (event.getItem() == null) {
    event.preventOpen();
  } else {
    // remove/add stuff to contextMenu
  }
});
```

Not the prettiest API, but I think it would be conceptually more correct. @Legioth, can you provide your opinion?

Legioth: The general convention in e.g. `Binder` is that there's an `addXyzListener` method in situations where multiple competing callbacks are supported and a `setXyzHandler` method if there can only be one. In this case, I think it makes most sense to only have one callback, which means that the "handler" concept might be appropriate.

Based on that point of view, my name suggestion without too much consideration would be `setDynamicContentHandler`. When removing "renderer" from the name, the words "dynamic" and "content" are again meaningful. I see the possibility of returning `false` more as an optimization rather than a primary feature, so I don't think the method name needs to take that into account.

pekam: So do I understand you correctly:
- You agree that "renderer" is a misleading term in this context
- It's okay to name this method based on dynamic content, because that's the use case it's created for, although the method is technically just a generic callback which doesn't enforce the user to use it for dynamic content handling

Legioth: - Yes
- Yes, the method can be named based on what it is intended to be used for rather than all the ways it can potentially be abused.

Legioth: Choosing to not show a menu at all is really "dynamic". Furthermore, it's a special form of "content" in the same way that the value of a list variable can either be an empty list or `null`.

pekam: Ok, that's fine for me. Thanks for your input!

So, @gatanaso, unless you disagree or want to discuss more, can you rename the methods to `setDynamicContentHandler` and `getDynamicContentHandler`?

gatanaso: @pekam Yes, I will update the methods accordingly.

PR #778, opened 2019-09-27
Add API for scrolling to index/start/end

### LINKED ISSUE #618: Add scroll to index API

tepi:
Fixes #618

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid-flow/778)

### DIFF
--- a/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
+++ b/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
@@ -121,7 +121,7 @@ import elemental.json.JsonValue;
  *
  */
 @Tag("vaadin-grid")
-@NpmPackage(value = "@vaadin/vaadin-grid", version = "5.5.0-alpha1")
+@NpmPackage(value = "@vaadin/vaadin-grid", version = "5.5.0-alpha4")
 @JsModule("@vaadin/vaadin-grid/src/vaadin-grid.js")
 @JsModule("@vaadin/vaadin-grid/src/vaadin-grid-column.js")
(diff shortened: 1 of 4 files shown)


vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
```diff
@@ -3801,4 +3801,33 @@ public void setColumnOrder(List<Column<T>> columns) {
...
+     */
+    public void scrollToStart() {
+        scrollToIndex(0);
+    }
+
+    /**
+     * Scrolls to the end of the last data row.
+     */
+    public void scrollToEnd() {
+        scrollToIndex(getElement().getProperty("size", 0));
```
tomivirkki: This might not work if a tree grid has items expanded. `size` is only the number of root-level items.

tepi: Can I use `_effectiveSize` then? Or do you have another suggestion?

tomivirkki: That should work

tepi: Had to do it a bit differently since for some reason getting the property value for `_effectiveSize` from the server side does not seem to work

tepi: Had to do it like this: `getElement().executeJs("$0.scrollToIndex($0._effectiveSize)", this.getElement());` since the `_effectiveSize` property does not seem to available from the server side.

PR #780, opened 2019-10-01
ItemClickListener.getColumn

### LINKED ISSUE #725: ItemClickListener add support for getting column from event
Sometimes it is handy to be able to determine which column on a grid row was clicked for instance: 

- for example a clickable link in the cell and you want to handle it on the server-side, even if you click next to the link. 
- Or you have a checkbox in a cell and want to toggle it, even if you do not directly click it

This was possible in the Vaadin 8 Grid itemClickEvent but is missing in the v14 itemClickEvent

Tested: 
Vaadin 14.0.0.rc6

vaadin-miki:
Added a method to ItemClickListener. It works by passing the id of the column to the event and then discovering the column on the server side.

I used the idea from the workaround by @pekam in https://vaadin.com/forum/thread/17585345/17586229 - but also used internal `flowId` of the column to get it when the user forgets to set the id.

Simple tests included.

Closes #725

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid-flow/780)

### DIFF
--- a/vaadin-grid-flow-demo/src/main/java/com/vaadin/flow/component/grid/demo/GridDemo.java
+++ b/vaadin-grid-flow-demo/src/main/java/com/vaadin/flow/component/grid/demo/GridDemo.java
@@ -1831,7 +1831,7 @@ public class GridDemo extends DemoView {
                 event -> {
                     name.setText(event.getItem().getFirstName());
                     age.setText(String.valueOf(event.getItem().getAge()));
-                    event.getColumn().ifPresent(col -> column.setText(col.getKey()));
+                    column.setText(event.getColumn().getKey());
                 });
 
         // end-source-example
--- a/vaadin-grid-flow-demo/src/main/java/com/vaadin/flow/component/grid/demo/GridDemo.java
+++ b/vaadin-grid-flow-demo/src/main/java/com/vaadin/flow/component/grid/demo/GridDemo.java
@@ -1810,13 +1810,14 @@ public class GridDemo extends DemoView {
         FormLayout formLayout = new FormLayout();
         Label name = new Label();
         Label age = new Label();
+        Label column = new Label();
 
         // begin-source-example
         // source-example-heading: Item Click Listener
         Grid<Person> grid = new Grid<>();
         grid.setItems(getItems());
-        grid.addColumn(Person::getFirstName).setHeader("First Name");
-        grid.addColumn(Person::getAge).setHeader("Age");
+        grid.addColumn(Person::getFirstName).setHeader("First Name").setKey("First Name").setId("first name");
+        grid.addColumn(Person::getAge).setHeader("Age").setKey("Age");
 
         // Disable selection: will receive only click events instead
         grid.setSelectionMode(Grid.SelectionMode.NONE);
@@ -1824,11 +1825,18 @@ public class GridDemo extends DemoView {
         formLayout.add(name, age);
         formLayout.addFormItem(name, "Name");
         formLayout.addFormItem(age, "Age");
+        formLayout.addFormItem(column, "Column");
 
         grid.addItemClickListener(
                 event -> {
                     name.setText(event.getItem().getFirstName());
                     age.setText(String.valueOf(event.getItem().getAge()));
+                    if(event.getColumn().isPresent()) {
+                        column.setText(event.getColumn().get().getKey());
+                    }
+                    else {
+                        column.setText("(undefined column");
+                    }
                 });
 
         // end-source-example
--- a/vaadin-grid-flow-demo/src/main/java/com/vaadin/flow/component/grid/demo/GridDemo.java
+++ b/vaadin-grid-flow-demo/src/main/java/com/vaadin/flow/component/grid/demo/GridDemo.java
@@ -1816,7 +1816,7 @@ public class GridDemo extends DemoView {
         // source-example-heading: Item Click Listener
         Grid<Person> grid = new Grid<>();
         grid.setItems(getItems());
-        grid.addColumn(Person::getFirstName).setHeader("First Name").setKey("First Name").setId("first name");
+        grid.addColumn(Person::getFirstName).setHeader("First Name").setKey("First Name");
         grid.addColumn(Person::getAge).setHeader("Age").setKey("Age");
 
         // Disable selection: will receive only click events instead
@@ -1831,12 +1831,7 @@ public class GridDemo extends DemoView {
                 event -> {
                     name.setText(event.getItem().getFirstName());
                     age.setText(String.valueOf(event.getItem().getAge()));
-                    if(event.getColumn().isPresent()) {
-                        column.setText(event.getColumn().get().getKey());
-                    }
-                    else {
-                        column.setText("(undefined column");
-                    }
+                    event.getColumn().ifPresent(col -> column.setText(event.getColumn().get().getKey()));
                 });
 
         // end-source-example
--- a/vaadin-grid-flow-demo/src/main/java/com/vaadin/flow/component/grid/demo/GridDemo.java
+++ b/vaadin-grid-flow-demo/src/main/java/com/vaadin/flow/component/grid/demo/GridDemo.java
@@ -1831,7 +1831,7 @@ public class GridDemo extends DemoView {
                 event -> {
                     name.setText(event.getItem().getFirstName());
                     age.setText(String.valueOf(event.getItem().getAge()));
-                    event.getColumn().ifPresent(col -> column.setText(event.getColumn().get().getKey()));
+                    event.getColumn().ifPresent(col -> column.setText(col.getKey()));
                 });
 
         // end-source-example
--- a/vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/it/ItemClickListenerPage.java
+++ b/vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/it/ItemClickListenerPage.java
@@ -17,6 +17,7 @@ package com.vaadin.flow.component.grid.it;
 
 import com.vaadin.flow.component.checkbox.Checkbox;
 import com.vaadin.flow.component.grid.Grid;
+import com.vaadin.flow.component.grid.ItemClickEvent;
 import com.vaadin.flow.component.html.Div;
 import com.vaadin.flow.router.Route;
 
@@ -30,17 +31,35 @@ public class ItemClickListenerPage extends Div {
         Div dblClickMsg = new Div();
         dblClickMsg.setId("dblClickMsg");
 
+        Div columnKeyMsg = new Div();
+        columnKeyMsg.setId("columnClickMsg");
+
         Grid<String> grid = new Grid<>();
         grid.setItems("foo", "bar");
-        grid.addColumn(item -> item).setHeader("Name");
+        grid.addColumn(item -> item).setHeader("Name").setKey("Name");
         grid.addComponentColumn(item -> new Checkbox(item));
 
-        grid.addItemClickListener(event -> clickMsg.add(event.getItem()));
+        grid.addItemClickListener(event -> {
+            clickMsg.add(event.getItem());
+            columnKeyMsg.add(getColumnKeyFromEvent(event));
+        });
 
-        grid.addItemDoubleClickListener(event -> dblClickMsg
-                .setText(String.valueOf(event.getClientY())));
+        grid.addItemDoubleClickListener(event -> {
+            dblClickMsg
+                    .setText(String.valueOf(event.getClientY()));
+            columnKeyMsg.setText(getColumnKeyFromEvent(event));
+        });
 
         add(grid, clickMsg, dblClickMsg);
+    }
+
+    private static String getColumnKeyFromEvent(ItemClickEvent<?> event) {
+        if(event.getColumn().isPresent()) {
+            return event.getColumn().get().getKey();
+        }
+        else {
+            return "(cannot get clicked column key!)";
+        }
     }
 
 }
--- a/vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/it/ItemClickListenerPage.java
+++ b/vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/it/ItemClickListenerPage.java
@@ -50,9 +50,8 @@ public class ItemClickListenerPage extends Div {
         });
 
         grid.addItemDoubleClickListener(event -> {
-            dblClickMsg
-                    .setText(String.valueOf(event.getClientY()));
-            columnDblClickMsg.setText(event.getColumn().getKey());
+            dblClickMsg.add(String.valueOf(event.getClientY()));
+            columnDblClickMsg.add(event.getColumn().getKey());
         });
 
         grid.setItemDetailsRenderer(new ComponentRenderer<>((SerializableFunction<String, Span>) ItemClickListenerPage::getDetailsComponent));
--- a/vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/it/ItemClickListenerPage.java
+++ b/vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/it/ItemClickListenerPage.java
@@ -17,7 +17,6 @@ package com.vaadin.flow.component.grid.it;
 
 import com.vaadin.flow.component.checkbox.Checkbox;
 import com.vaadin.flow.component.grid.Grid;
-import com.vaadin.flow.component.grid.ItemClickEvent;
 import com.vaadin.flow.component.html.Div;
 import com.vaadin.flow.component.html.Span;
 import com.vaadin.flow.data.renderer.ComponentRenderer;
@@ -43,17 +42,17 @@ public class ItemClickListenerPage extends Div {
         Grid<String> grid = new Grid<>();
         grid.setItems("foo", "bar");
         grid.addColumn(item -> item).setHeader("Name").setKey("Name");
-        grid.addComponentColumn(item -> new Checkbox(item));
+        grid.addComponentColumn(Checkbox::new);
 
         grid.addItemClickListener(event -> {
             clickMsg.add(event.getItem());
-            columnClickMsg.add(getColumnKeyFromEvent(event));
+            columnClickMsg.add(event.getColumn().getKey());
         });
 
         grid.addItemDoubleClickListener(event -> {
             dblClickMsg
                     .setText(String.valueOf(event.getClientY()));
-            columnDblClickMsg.setText(getColumnKeyFromEvent(event));
+            columnDblClickMsg.setText(event.getColumn().getKey());
         });
 
         grid.setItemDetailsRenderer(new ComponentRenderer<>((SerializableFunction<String, Span>) ItemClickListenerPage::getDetailsComponent));
@@ -67,15 +66,6 @@ public class ItemClickListenerPage extends Div {
         Span result = new Span(s);
         result.setId("details-"+s);
         return result;
-    }
-
-    private static String getColumnKeyFromEvent(ItemClickEvent<?> event) {
-        if (event.getColumn().isPresent()) {
-            return event.getColumn().get().getKey();
-        }
-        else {
-            return "(cannot get clicked column key!)";
-        }
     }
 
 }
--- a/vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/it/ItemClickListenerPage.java
+++ b/vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/it/ItemClickListenerPage.java
@@ -50,7 +50,7 @@ public class ItemClickListenerPage extends Div {
             columnKeyMsg.setText(getColumnKeyFromEvent(event));
         });
 
-        add(grid, clickMsg, dblClickMsg);
+        add(grid, clickMsg, dblClickMsg, columnKeyMsg);
     }
 
     private static String getColumnKeyFromEvent(ItemClickEvent<?> event) {
--- a/vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/it/ItemClickListenerPage.java
+++ b/vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/it/ItemClickListenerPage.java
@@ -31,8 +31,11 @@ public class ItemClickListenerPage extends Div {
         Div dblClickMsg = new Div();
         dblClickMsg.setId("dblClickMsg");
 
-        Div columnKeyMsg = new Div();
-        columnKeyMsg.setId("columnClickMsg");
+        Div columnClickMsg = new Div();
+        columnClickMsg.setId("columnClickMsg");
+
+        Div columnDblClickMsg = new Div();
+        columnDblClickMsg.setId("columnDblClickMsg");
 
         Grid<String> grid = new Grid<>();
         grid.setItems("foo", "bar");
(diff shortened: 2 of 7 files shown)


vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/ItemClickEvent.java
```diff
@@ -90,6 +96,18 @@ public ItemClickEvent(Grid<T> source, boolean fromClient,
         super(source, fromClient, screenX, screenY, clientX, clientY,
                 clickCount, button, ctrlKey, shiftKey, altKey, metaKey);
         item = source.getDataCommunicator().getKeyMapper().get(itemKey);
+        // id of column should be set for this to work
+        if(columnKey != null && !columnKey.isEmpty()) {
+            column = source.getColumns().stream().filter(col -> col.getId().isPresent() && columnKey.equals(col.getId().get())).findFirst().orElse(null);
+        }
+        // if id is not set, use internal flow ids
+        else if(flowColumnKey != null && !flowColumnKey.isEmpty() && flowColumnKey.matches("col\\d+")) {
```
pekam: Using the `_flowId` should be a reliable way to retrieve the column instance at server-side, so it should be the only method to use, instead of being a backup-solution in case the user didn't set the id-attributes.

Grid creates these unique ids for all columns automatically, and saves the references in `idToColumnMap`: https://github.com/vaadin/vaadin-grid-flow/blob/61b6c67afc25bc39511ea1e786d8c322fb10471f/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java#L1605-L1606

This `idToColumnMap` should be used here to retrieve the `Column` instance based on the column id. The problem is that it's private. I suggest adding a package-protected method to Grid:
`Column<T> getColumnByInternalId(String id)`
to retrieve values from that map.

Note: It's important to distinguish the terms "id" and "key" in this context. Ids are part of the Grid's internal implementation, and keys are the identifiers that the user can define. Also note that I'm not talking about HTML id-attribute (`setId`) here. :)

Another note: `_flowId` is the name of the property to set for the `<vaadin-grid-column>` web component. It's named with "flow"-prefix just to point that it's part of the Flow component implementation, and to avoid conflicting with any of the web component's own properties. At server-side, the "flow"-part is irrelevant in naming, as it's obviously running Flow. :)

vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/ItemClickEvent.java
```diff
@@ -77,6 +81,8 @@
      */
     public ItemClickEvent(Grid<T> source, boolean fromClient,
             @EventData("event.detail.itemKey") String itemKey,
+            @EventData("event.detail.columnKey") String columnKey,
+            @EventData("event.detail.flowKey") String flowColumnKey,
```
pekam: Technically adding these args is a breaking change, because it will remove the constructor with the earlier set of arguments. Although these event objects should be constructed by the framework, it's still possible for users to call the constructor.

I'd say that we have to add another constructor with the previous set of arguments, but without the `@EventDatas` to keep backwards compatibility, and hope that Flow is smart enough to pick this constructor with the `@EventDatas`.

@Legioth, any advice for this kind of situation: a new feature which requires adding `@EventData`?

vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
```diff
@@ -922,6 +922,7 @@ window.Vaadin.Flow.gridConnector = {
 
     grid.addEventListener('cell-activate', e => {
       grid.$connector.activeItem = e.detail.model.item;
+      e
```
pekam: This looks like an accident

vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
```diff
@@ -944,6 +945,13 @@ window.Vaadin.Flow.gridConnector = {
     function _fireClickEvent(event, eventName) {
       if (grid.$connector.activeItem) {
         event.itemKey = grid.$connector.activeItem.key;
+        const eventContext = grid.getEventContext(event);
+        if (eventContext.column.id) {
+          event.columnKey = eventContext.column.id;
+        }
+        // flowId should be always set and will be used as
+        // backup when the user forgets about setting column's id
+        event.flowKey = eventContext.column._flowId;
```
pekam: If you have a details-renderer, and click the details row, it should fire an `ItemClickEvent`, but `getEventContext().column` is undefined. So this needs a falsy-check.
This should also have a test-case.

vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/ItemClickEvent.java
vaadin-miki: @pekam @Legioth since since the first parameter is a string, is it acceptable to pass the flow id of the column in it? I would use some magic connector, like `::` to split it into real item key and column identifier? if `::` is not found, it means there is no column id (backwards compatibility) - just hoping there is no way to have item key itself include `::`, is that possible?

vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
vaadin-miki: in this case, is it possible to obtain the clicked column some other way?

vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
vaadin-miki: my entire IT career looks like an accident :P

vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/ItemClickEvent.java
vaadin-miki: ok, that I did not know, thanks :) code updated to follow the above

vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
pekam: No, because the click doesn't target any column in the case of details cell, which spans through the whole grid. That's why `getEventContext` has been designed to not have a column-reference in this case.

vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
```diff
@@ -944,6 +944,11 @@ window.Vaadin.Flow.gridConnector = {
     function _fireClickEvent(event, eventName) {
       if (grid.$connector.activeItem) {
         event.itemKey = grid.$connector.activeItem.key;
+        const eventContext = grid.getEventContext(event);
+        // if you have a details-renderer, getEventContext().column is undefined
+        if (eventContext.column) {
+          event.flowKey = eventContext.column._flowId;
```
pekam: Just an internal naming thing, but I'd like to see all variables referring to the internal column id be called `columnId` instead of `flowKey`, `flowColumnKey` etc. This makes it easier to read the code and understand that it's the thing that is used in `idToColumnMap`. Using "key" can cause confusion, since column-keys are the things that can be set by user. And like I mentioned earlier, the "flow"-part is used just for the web component's property name to avoid naming conflicts in the wild west of JavaScript.

vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/ItemClickEvent.java
```diff
@@ -101,4 +107,15 @@ public T getItem() {
         return item;
     }
 
+    /**
+     * Gets the column that was clicked.
+     * For this to really work the column's id must be set. Otherwise internal magic will be used
+     * to figure out which column it is (using column's internal flow identifiers that are then
+     * converted to integer index of the column).
+     * @return An optional with column. Optional may be empty if the column could not be figured out.
```
pekam: JavaDocs should be updated.

It should say that the optional is empty when hitting a details-cell.

vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/ItemClickEvent.java
pekam: That's quite a clever idea, and it should work because `itemKey` is an auto-generated positive integer. But that's also really hacky and I'd like to avoid that.

I still think that another constructor would be the best solution, but would like to wait for @Legioth's opinion, since this is the first time for me to see this kind of situation.

vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/ItemClickEvent.java
vaadin-miki: Done.

vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
vaadin-miki: I used `flowId` - `columnId` sounds to me like a thing you can set with `column.setId`, where as `flowId` indicates it is Flow's internal. Also change the new method in Grid to be `getColumnByFlowId` to make it more consistent. Does this make sense?

vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
pekam: It still needs a test case (clicking a details cell).

vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
pekam: Well, I still prefer my idea. I might be biased because I'm used to working with Grid internals and I know there are these internal ids, so I don't get confused with `column.setId`.

If you'd look for the first time at the arguments of `new ItemClickEvent()` and see there `String flowId`, would you understand based on this name what it is? Id of Flow framework? There's nothing in that name indicating that it's actually column-specific identifier.

I think the term "internal" is key here, not "flow". So if `columnId` gets confused with `column.setId`, IMO the best option would be `columnInternalId` or `internalColumnId`.

vaadin-miki: `internalColumnId` it is then :)

vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/ItemClickListenerIT.java
```diff
@@ -79,6 +79,32 @@ public void doubleClickCheckboxInCell_noEventsFired() {
...
+    }
+
+    @Test
+    public void clickComponentInCell_columnNameAvailable() {
+        grid.getCell(0, 0).click();
+        TestBenchElement checkbox = grid.getCell(0, 1).$("vaadin-checkbox")
+                .first();
+        checkbox.click();
+        Assert.assertEquals("Name", getColumnClickMessage());
+    }
```
pekam: The `ItemClickEvent` is specifically implemented in such a way that it doesn't fire when clicking focusable components, such as checkbox.

I'm not sure what is the purpose of this test. It's name implies that clicking the checkbox should fire event, but actually it doesn't. The only event comes from clicking the cell on the first column. This is already covered in the previous test.

There is already a test case for `clickCheckboxInCell_noItemClickEventFired`, so I think this case can be just removed.

vaadin-grid-flow-demo/src/main/java/com/vaadin/flow/component/grid/demo/GridDemo.java
```diff
@@ -1810,25 +1810,33 @@ private void createClickListener() {
...
         grid.addItemClickListener(
                 event -> {
                     name.setText(event.getItem().getFirstName());
                     age.setText(String.valueOf(event.getItem().getAge()));
+                    if(event.getColumn().isPresent()) {
+                        column.setText(event.getColumn().get().getKey());
+                    }
+                    else {
+                        column.setText("(undefined column");
+                    }
```
pekam: 1. There shouldn't be a situation where the column is not defined, so the else-part could be just removed.

2. Could be cleaner this way:
`event.getColumn().ifPresent(col -> column.setText(col.getKey()))`
(not blocking)

3. Our code style rules state that there should be a space after `if`. Would be better to fix elsewhere as well, but blocking only for this demo snippet which is shown to users.

vaadin-grid-flow-demo/src/main/java/com/vaadin/flow/component/grid/demo/GridDemo.java
```diff
@@ -1810,25 +1810,33 @@ private void createClickListener() {
...
         Label age = new Label();
+        Label column = new Label();
 
         // begin-source-example
         // source-example-heading: Item Click Listener
         Grid<Person> grid = new Grid<>();
         grid.setItems(getItems());
-        grid.addColumn(Person::getFirstName).setHeader("First Name");
-        grid.addColumn(Person::getAge).setHeader("Age");
+        grid.addColumn(Person::getFirstName).setHeader("First Name").setKey("First Name").setId("first name");
```
pekam: I guess `setId` could be removed now to keep the demo simpler.

vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/ItemClickListenerIT.java
```diff
@@ -79,6 +79,32 @@ public void doubleClickCheckboxInCell_noEventsFired() {
...
+
+    @Test
+    public void doubleClickCell_columnNameAvailable() {
+        grid.getCell(0, 0).doubleClick();
+        Assert.assertEquals("Name", getColumnClickMessage());
```
pekam: This is kind of ugly, because it depends on the click-listener adding the "Name"-text twice before the double-click event is fired, which sets the value to single "Name". Would be cleaner to have separate messages for click and dblclick.

pekam: It could turn into a false-positive in many ways, e.g.:
1. For some reason, `dblclick` gets fired before the second `click` on some browser.
2. Someone is not careful and changes the test-page so that `click` also calls `setText("name")` instead of adding text.

vaadin-grid-flow-demo/src/main/java/com/vaadin/flow/component/grid/demo/GridDemo.java
vaadin-miki: Done.

vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/ItemClickListenerIT.java
vaadin-miki: Agree. Didn't fully understand how tests work here, my fault. Fixed.

vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
vaadin-miki: what would be the correct thing to do, because all have pros and cons:
1. build a separate IT test for that
2. check it in e.g. `GridDetailsRowIT`
3. modify existing `ItemClickListenerIT` to provide details for the row

vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/ItemClickListenerIT.java
```diff
@@ -63,22 +63,34 @@ public void clickCheckboxInCell_noItemClickEventFired() {
         Assert.assertEquals("", getClickMessage());
     }
 
-    @Test
-    public void clickCell_clickCheckboxInCell_onlyOneClickEventFired() {
```
pekam: Why did you remove an existing test?

vaadin-grid-flow-demo/src/main/java/com/vaadin/flow/component/grid/demo/GridDemo.java
```diff
@@ -1810,25 +1810,28 @@ private void createClickListener() {
...
         formLayout.add(name, age);
         formLayout.addFormItem(name, "Name");
         formLayout.addFormItem(age, "Age");
+        formLayout.addFormItem(column, "Column");
 
         grid.addItemClickListener(
                 event -> {
                     name.setText(event.getItem().getFirstName());
                     age.setText(String.valueOf(event.getItem().getAge()));
+                    event.getColumn().ifPresent(col -> column.setText(event.getColumn().get().getKey()));
```
pekam: Use the lambda parameter: `event.getColumn().get()` -> `col`

vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
pekam: 3. would be most appropriate IMO

vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/ItemClickListenerIT.java
vaadin-miki: not my day :| will bring it back with the next commit... *sigh* good there are code reviews

vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/ItemClickListenerIT.java
```diff
@@ -79,6 +88,15 @@ public void clickCell_columnNameAvailable() {
         Assert.assertEquals("Name", getColumnClickMessage());
     }
 
+    @Test
+    public void clickComponentInCell_columnNameAvailable() {
```
pekam: This was correctly removed in previous commit and now you brought it back. :)

vaadin-miki: You know what, I will get back to this tomorrow. I am afraid this will not go anywhere good today.

vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/ItemClickEvent.java
Legioth: You can leave the old constructor signature without any `@EventData` annotations and then add a new one with all of them. At least based on reading the code in `com.vaadin.flow.component.ComponentEventBusUtil.findEventConstructor(Class<T>)`, the constructor without annotated parameter should simply be ignored.

vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/ItemClickListenerIT.java
```diff
@@ -79,6 +80,37 @@ public void doubleClickCheckboxInCell_noEventsFired() {
...
+    }
+
+    @Test
+    public void doubleClickCell_columnNameAvailable() {
+        grid.getCell(0, 0).doubleClick();
+        Assert.assertEquals("Name", getColumnDoubleClickMessage());
+    }
+
+    @Test
+    public void clickDetailsCell_noColumnNameAvailable() {
```
pekam: Great! We should also have a test case that `getItem()` works when clicking details cell. But that is not part of this feature ,so it's up to you if you want to add that test case as well.

pekam: Oh man... Looks like I was wrong all the time. Clicking on details-cell doesn't fire the `item-click` event at all. Sorry for that. :/
It's because clicking on details-cell doesn't fire the web component's `cell-activate` event, which is used to determine the target item.
So can you change this test case:
- Change the name to `clickDetailsCell_noItemClickEventFired`
- Check also that there's no item name logged (in addition to column)

vaadin-miki: @pekam because item click is not fired, does this mean there is no way for column information to be missing? if so, I could get rid of the `Optional` in event and just return the column

pekam: Good point! I think so, yes. At least there _shouldn't_ be a case where the column is null.
I try to think/test if there are any corner cases where that could happen...

vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/it/ItemClickListenerPage.java
```diff
@@ -30,17 +33,39 @@ public ItemClickListenerPage() {
...
+        grid.addItemClickListener(event -> {
+            clickMsg.add(event.getItem());
+            columnClickMsg.add(event.getColumn().getKey());
+        });
 
-        grid.addItemClickListener(event -> clickMsg.add(event.getItem()));
+        grid.addItemDoubleClickListener(event -> {
+            dblClickMsg
+                    .setText(String.valueOf(event.getClientY()));
+            columnDblClickMsg.setText(event.getColumn().getKey());
```
pekam: One more thing here: There's a reason why writing to the message used `add` instead of `setText`. This way it catches bugs where duplicate events are fired, because in that case e.g. `columnDblClickMsg` would become "NameName" instead of "Name", and the test would fail. And I do remember actually seeing this kind of duplicate-event bugs before. So please change these two from `setText` to `add`, to make the test more robust.

vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/ItemClickListenerIT.java
```diff
@@ -79,6 +80,39 @@ public void doubleClickCheckboxInCell_noEventsFired() {
...
+        grid.getCell(0, 0).doubleClick();
+        Assert.assertEquals("Name", getColumnDoubleClickMessage());
+    }
+
+    @Test
+    public void clickDetailsCell_noItemClickEventFired() {
+        waitUntil(driver ->
+                      grid.findElements(By.className("row-details")) != null);
+        WebElement details = findElement(By.id("details-bar"));
+        details.click();
```
pekam: Could also do a `doubleClick()` to verify that one also. Adding it here is fine IMO. Or you can create a separate test case if you wish.

vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/ItemClickListenerIT.java
pekam: Couldn't find any way to break it :)

vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/ItemClickEvent.java
```diff
@@ -101,4 +158,13 @@ public T getItem() {
         return item;
     }
 
+    /**
+     * Gets the column that was clicked.
+     * @return The column that was clicked. It should not be {@code null}, unless somehow the information
+     * about the column has not been passed from the client-side component.
```
pekam: Formatting:
- empty line before `@return`
- no complete sentences in `@return`

>It should not be {@code null}, unless somehow....


^This is pretty un-conventional documentation. It's basically saying that it shouldn't be null, but there might be a bug which we're not aware of. The same applies to all of the APIs we have. :D
Change to e.g. "the column that was clicked, not {@code null}"

vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/ItemClickEvent.java
```diff
@@ -45,6 +47,9 @@
      *            client, <code>false</code> if the event originates from
      *            server-side logic
      * @param itemKey the item mapper key
+     * @param internalColumnId
+     *            the internal id of the column associated with
+     *            the click event (if present)
```
pekam: remove "(if present)"

vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/ItemDoubleClickEvent.java
```diff
@@ -43,6 +43,9 @@
      *            client, <code>false</code> if the event originates from
      *            server-side logic
      * @param itemKey the item mapper key
+     * @param internalColumnId
+     *            the internal id of the column associated with
+     *            the click event (if present)
```
pekam: remove "(if present)"

PR #781, opened 2019-10-02
Implement setDeselectAllowed

### LINKED ISSUE #713: setDeselectAllowed(false) does not work as expected
Using vaadin.version 14.0.0.rc4, I am not seeing the expected behaviour when disallowing deselection of a grid item. 

```
    grid = new Grid<>(Person.class, false);
    grid.setItems(getData(..));

    grid.setSelectionMode(Grid.SelectionMode.SINGLE);

    // does not work as expected
    ((GridSingleSelectionModel) grid.getSelectionModel()).setDeselectAllowed(false);
```

Expected behaviour: Not being able to unselect a (pre-selected) grid item, disallowing zero selected items on the grid.

Actual behaviour: Still being able to deselect (and select again) a grid item.

tepi:
Fixes #713

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid-flow/781)

### DIFF
--- a/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/GridSingleSelectionIT.java
+++ b/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/GridSingleSelectionIT.java
@@ -1,0 +1,88 @@
+/*
+ * Copyright 2000-2019 Vaadin Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.vaadin.flow.component.grid.it;
+
+import org.junit.Assert;
+import org.junit.Test;
+
+import com.vaadin.flow.component.grid.testbench.GridElement;
+import com.vaadin.flow.testutil.AbstractComponentIT;
+import com.vaadin.flow.testutil.TestPath;
+
+@TestPath("grid-single-selection")
+public class GridSingleSelectionIT extends AbstractComponentIT {
+
+    @Test
+    public void checkDeselectionAllowedByDefault() {
+        open();
+
+        // Ensure that de-selection is allowed by default
+        GridElement grid = $(GridElement.class)
+                .id(GridSingleSelectionPage.DESELECT_ALLOWED_GRID_ID);
+
+        grid.getRow(1).select();
+        Assert.assertTrue("Row 1 was not selected after selecting it.",
+                grid.getRow(1).isSelected());
+
+        grid.getRow(1).deselect();
+        Assert.assertTrue("Row 1 was still selected after de-selecting it.",
+                !grid.getRow(1).isSelected());
+
+        // Disable de-selection on the fly and test again
+        $("button").id(GridSingleSelectionPage.DESELECT_ALLOWED_TOGGLE_ID)
+                .click();
+
+        grid.getRow(1).select();
+        Assert.assertTrue("Row 1 was not selected after selecting it.",
+                grid.getRow(1).isSelected());
+
+        grid.getRow(1).deselect();
+        Assert.assertTrue(
+                "Row 1 was deselected even though deselection is not allowed.",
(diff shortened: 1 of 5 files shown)


vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
```diff
@@ -224,6 +224,14 @@ protected void fireSelectionEvent(
...
                         grid.fireEvent((ComponentEvent<Grid<T>>) event);
                     }
+
+                    @Override
+                    public void setDeselectAllowed(boolean deselectAllowed) {
+                        super.setDeselectAllowed(deselectAllowed);
+                        grid.getElement().executeJs(
+                                "this.$connector.deselectAllowed = $0",
+                                deselectAllowed);
+                    }
```
pekam: As the deselection is prevented only in the client-side, this is currently a pure UX feature, which can be hacked by running `grid.deselectAllowed = true` in the browser.

As a developer, I might expect that after calling `setDeselectAllowed(false)`, the value can't change to `null`.
If the developer is not cautious and handle the value change event with `null`, a hackerman can break the app.

The JavaDocs states "Sets whether it's allowed to deselect the selected row through UI", which kind-of indicates that it should be more of an UX-feature. Still, I think that it would be better to prevent deselection also at the server-side to avoid security issues.

pekam: As discussed f2f, adding a check [here](https://github.com/vaadin/vaadin-grid-flow/blob/ca8cb26b9de724b0f4579067daceba94773800ab/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/AbstractGridSingleSelectionModel.java#L88) would probably be the best way to fix it so that would still allow deselection with the Java APIs

tepi: Added check

pekam: It should be actually without `!`. Currently the deselection is broken, as it doesn't update the server-side state to `null` value when deselection is allowed.

tepi: True that. Fixed.

vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/GridSingleSelectionIT.java
```diff
@@ -0,0 +1,88 @@
...
+
+import com.vaadin.flow.component.grid.testbench.GridElement;
+import com.vaadin.flow.testutil.AbstractComponentIT;
+import com.vaadin.flow.testutil.TestPath;
+
+@TestPath("grid-single-selection")
+public class GridSingleSelectionIT extends AbstractComponentIT {
+
+    @Test
+    public void checkDeselectionAllowedByDefault() {
```
pekam: "By default" here and in the comment below is incorrect. Because the `buildGrid` method calls `setDeselectAllowed` for both Grids, there is no test case ensuring how it works by default. Currently, these tests don't ensure that it works correctly without calling the setter.

I suggest not to call that setter in `buildGrid` for this first Grid.

vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
```diff
@@ -199,15 +199,13 @@ window.Vaadin.Flow.gridConnector = {
...
-        if (oldVal && selectedKeys[oldVal.key]) {
+    	if (!grid.$connector.deselectAllowed && oldVal) {
+    	  grid.activeItem = oldVal;
+    	} else if (oldVal && selectedKeys[oldVal.key]) {
           grid.$connector.doDeselection([oldVal], true);
         }
-        return;
-      }
-      if (!selectedKeys[newVal.key]) {
+      } else if (!selectedKeys[newVal.key]) {
```
pekam: There are some tabs in the indentation. Change to spaces.

pekam:
Also, please create new commits instead of force pushing, so we can easily see the changes since last revision. We will squash-merge this so the history in this branch doesn't need to look pretty. :)

vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/GridSingleSelectionIT.java
tepi: Fixed

vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
tepi: Fixed

PR #786, opened 2019-10-09
Fix column visual reorder

tomivirkki:
Fixes a DX test finding: "If the columns are already physically in correct order in the dom, the column reorder doesn't have any effect even if the columns are visually in different order"

Note that the first commit in this PR is just about autoformatted lines. The meaningful changes are in the second commit.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid-flow/786)

### DIFF
--- a/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/GridOrderColumnsIT.java
+++ b/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/GridOrderColumnsIT.java
@@ -64,6 +64,14 @@ public class GridOrderColumnsIT extends AbstractComponentIT {
         assertColumnHeaders("Col1", "Col2", "Col3");
     }
 
+    @Test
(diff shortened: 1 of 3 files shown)


vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/GridOrderColumnsIT.java
```diff
@@ -64,6 +64,14 @@ public void gridOrder_321_123() {
         assertColumnHeaders("Col1", "Col2", "Col3");
     }
 
+    @Test
+    public void gridVisualOrderReset() {
```
pekam: I have no idea based on the test name what is the case that is actually covered.
I'd prefer something like
`clientReorderColumns_serverResetOrder`

Not blocking.

tomivirkki: Done.

Issue #366, opened 2018-10-10
Click Listener for middle or right mouse buttons


F43nd1r:
Right now the click listener introduced in #359 (and released in 2.1.0.alpha1) only works for left clicks. It would be nice to support the other mouse buttons as well.

gilberto-torrezan:
Confirmed. The API has support for multiple buttons, but the event never reaches the server. For example, in this code (based on the [GridView](https://github.com/vaadin/vaadin-grid-flow/blob/8bd81095f13aebee77c047d719abcb2408ea3da8/src/test/java/com/vaadin/flow/component/grid/demo/GridView.java#L1298-L1320) demo), the cases 1 and 2 are never executed:

```java
Div message = new Div();
Grid<Person> grid = new Grid<>();
grid.setItems(getItems());
grid.addColumn(Person::getName).setHeader("Name");
grid.addColumn(Person::getAge).setHeader("Age");

grid.setSelectionMode(SelectionMode.NONE);

grid.addItemClickListener(event -> {
            switch (event.getButton()) {
            case 0:
                message.setText("Clicked Item: " + event.getItem());
                break;
            case 1:
                message.setText("Middle-Clicked Item: " + event.getItem()); // never executed
                break;
            case 2:
                message.setText("Right-Clicked Item: " + event.getItem()); // never executed
                break;
            }
        });
```

Apparently the Grid connector needs to prevent the default click behavior on the browser in order for those events to be triggered.

pleku:
For the case *when the item click listener is added* on the server side, the defaults should be:
- the event bubbling / propagated is stopped
- the default (browser behavior) is prevented

Right and middle clicks should obviously work.

pleku:
Just for reference, we have https://github.com/vaadin/flow/issues/1363 for making it possible to give control for the developer to decide what to do with bubbling and preventing default.

sebdavid:
Hi,
Do you have a plan about the fix and release of this issue?
Or do you have some kind of workaround?

My use case would be that by right-clicking an item of the grid, it would be automatically selected and the contextmenu opened.

sebdavid:
Hi,
Actually, it seems that the right-clicking issue is a general Vaadin issue, I mean it it not bound to vaadin-grid-flow. 
The following code, directly applied on a VerticalLayout, does trigger an event only with a left-click :
```
addListener(ClickEvent.class, e -> {
    System.out.println("clicking :"+ e.getButton());
});
```

[...]

PR #818, opened 2019-11-07
Fix for #366

stefanuebe:
Issue fix for #366 

The `gridConnector` now adds listeners to the grid for `mouseup`, `mousedown` and `contextmenu` events. 

The `contextmenu` and the `mouseup` event lead finally to `_fireClickEventOnOtherMouseButtons` which simulates partially the normal click handling, without activating the item or selecting it. Also depending on set properties, the default browser actions are prevented (see details below).

The `mousedown` event is needed for the special case of preventing the browser default action of the middle key (mostly scrolling). 

All events lead to the `ItemClickEvent` in Java and allow usage of `event.getButton()`. 

Also the grid now provides a method `preventDefaultMouseClickActions(boolean)`, so that the dev can decide to prevent the default actions, that inflict on middle our right mouse button. (This method is independent from the click listener).

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid-flow/818)


tomivirkki:
Would this fire a row click event on Android if you press a row long enough to create a "contextmenu" event? This would be unexpected as it's not a click and it might behave differently on iOS.

stefanuebe:
Tested the long touch on a Surface Book Windows Chrome, Firefox and Edge and on Android  Chrome. In all cases the long click only fired the `contextmenu` event, not the normal click event. 

Firefox and Edge send `button 2` with the mouse event. Chrome sends `button 0` (Windows and Chrome), which is a problem, since that indicates on the server side, that the user clicked "normally" on the item. 

Question is, if we want to modify the mouse event in the case of Chrome and set the `button` manually to `2`?

stefanuebe:
Added above described special case handling

tomivirkki:
> In all cases the long click only fired the contextmenu event, not the normal click event

What I meant is that a listener added with addItemClickListener shouldn't invoke on row long press on touch devices because it's not a click and it doesn't work consistently across browsers (iOS Safari). On desktop, it also might be strange if a _click_ listener is invoked when the "contextmenu" event occurs by "pressing the context menu key" as described here https://developer.mozilla.org/en-US/docs/Web/API/Element/contextmenu_event.

tomivirkki:
Closing this. I think the feature needs more design. As discussed, a long touch, which causes a contextmenu-event is not really a click-event. Maybe a dedicated listener for contextmenu -event makes more sense. But in every case, this would count as a new feature.

Issue #839, opened 2019-12-04
TreeGrid with HierarchyColumn is not serializable


mas4ivv:
When adding a hierarchy column to a TreeGrid, it is no longer serializable. Simple test case:

    @Test
    public void serializeTreeGrid() throws IOException {
        TreeGrid<String> grid = new TreeGrid<>();
        grid.addHierarchyColumn(String::toString);
        ObjectOutputStream out = new ObjectOutputStream(new ByteArrayOutputStream());
        out.writeObject(grid);
        ....

This will result in an exception:
java.io.NotSerializableException: com.vaadin.flow.component.treegrid.TreeGrid$$Lambda$414/840109665
...

Haprog:
This fix has been included in vaadin-grid-flow [4.1.1](https://github.com/vaadin/vaadin-grid-flow/releases/tag/4.1.1)

PR #840, opened 2019-12-04
Fixing setComparator to use SerializableComparator

TatuLund:
Fixes https://github.com/vaadin/vaadin-grid-flow/issues/839


TatuLund:
There is still problems with static methods of java.util.Comparator. Those are not inherited by SerializableComparator. So completing this fix would require to solve 
https://github.com/vaadin/flow/issues/6226 first.

However in case of Comparator this is not straight forward, since java.util.Comparators is package private and the static methods are short hands to methods defined there.

TatuLund:
Closed for now, see #841 instead

PR #859, opened 2019-12-27
Add condition to check items contain elements are null in callSelecti…

### LINKED ISSUE #850: Grid: setDeselectAllowed (false) not usable because of an exception
Version: 14.1.2 (observed just after upgrade) 
Description: If we define a grid, one of the simplest and largely used sequence of operations raise an exception:   
- set grid with deselected allowed false (relatively new vaadin feature)
- grid: set items 
- grid: select first item 
- grid: set items (again)
The exception is not raised for deselected allowed false or if we do not select any grid row.

In all our modules (many). now we have bugs in all the basic operations (search + search) or (search + clear), where the above described sequence is used. It is a very high severity bug: the new feature "deselected allowed false " cannot be used.  
Exception 
- java.lang.RuntimeException: java.lang.IllegalStateException: Unknown key: 1
- function Grid:: findByKey 

import java.util.ArrayList;
import java.util.List;

import com.vaadin.flow.component.grid.AbstractGridSingleSelectionModel;
import com.vaadin.flow.component.grid.Grid;
import com.vaadin.flow.component.html.Div;
import com.vaadin.flow.router.Route;

@SuppressWarnings("serial")
@Route("TestDeselectAllowed")
public class TestDeselectAllowed extends Div {

	Grid<Item> grid;

	/** here is the test call */
	TestDeselectAllowed() {

		setupTest();

		/** fail example */
		test_deselectNotAllowed_SetSelectSet_fail();

		/** examples were is not failing */
		// test_deselectAllowed_SetSelectSet_ok();
		// test_deselectNotAllowed_SetNoSelectSet_ok();
	}

	void setupTest() {
		grid = new Grid<>();
		grid.addColumn(Item::getCode);
		grid.addColumn(Item::getDescription);
		add(grid);
	}

	void test_deselectNotAllowed_SetSelectSet_fail() {

		setDeselectAllowed(grid, false);
		
		List<Item> list = Item.createList();
		grid.setItems(list);

		grid.getSelectionModel().select(list.get(0));

		// fill after select
		grid.setItems(Item.createList());

	}

	void test_deselectAllowed_SetSelectSet_ok() {

		setDeselectAllowed(grid, true);
		List<Item> list = Item.createList();
		grid.setItems(list);

		grid.getSelectionModel().select(list.get(0));

		// fill after select
		grid.setItems(Item.createList());
	}

	void test_deselectNotAllowed_SetNoSelectSet_ok() {
		setDeselectAllowed(grid, false);
		List<Item> list = Item.createList();
		grid.setItems(list);

		// no select
		grid.setItems(Item.createList());
	}

	static void setDeselectAllowed(Grid<?> grid, boolean deselectAllowed) {
		if (grid.getSelectionModel() instanceof AbstractGridSingleSelectionModel) {
			((AbstractGridSingleSelectionModel<?>) grid.getSelectionModel())
					.setDeselectAllowed(deselectAllowed);
		}
	}

	public static class Item {

		public String getCode() {
			return code;
		}

		public String getDescription() {
			return description;
		}

		String code;
		String description;

		public Item() {
			super();
		}

		public static List<Item> createList() {
			List<Item> list = new ArrayList<>();
			for (int i = 0; i < 5; i++) {
				Item item = new Item();
				item.code = "code " + i;
				item.description = "desc " + i;
				list.add(item);
			}

			return list;
		}
	}

}

hongduc-phan:
Fixed Grid: setDeselectAllowed (false) not usable because of an exception.

Fixes #850


vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
```diff
@@ -2532,9 +2532,25 @@ public void deselect(T item) {
      * @see GridSelectionModel
      */
     public void deselectAll() {
+        setDeselectAllowed(this,true);
```
tulioag: Space is missing after comma

vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
```diff
@@ -2532,9 +2532,25 @@ public void deselect(T item) {
...
 
+    /**
+     * This method is to set DeselectAllowed
+     *
+     * @param grid
+     *           the current grid
+     * @param deselectAllowed
+     *           value to set for deselectAllowed
+     */
+    private void setDeselectAllowed(Grid<?> grid, boolean deselectAllowed) {
```
tulioag: 1. Parameter grid is redundant, it will only be used with `this`.
1. The javadoc has no information that can't be deduced by reading the method declaration and is unnecessary for a private method.
1. Doesn't anything need to be done in multi selection?

vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
tulioag: After calling this method, the grid will allow deselection in the UI, even if it was not the case previously.

PR #860, opened 2020-01-03
Fix: grid flow set deselect allowed to false

### LINKED ISSUE #850: Grid: setDeselectAllowed (false) not usable because of an exception

hongduc-phan:
Add more new condition `selectedKeys[oldVal.key]` in `if (!grid.$connector.deselectAllowed && oldVal && selectedKeys[oldVal.key])`. The condition "cares" about when grid.activeItem = oldVal.

Fixes #850


vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/GridSingleSelectionIT.java
```diff
@@ -85,4 +85,25 @@ public void checkDeselectionDisallowedInitially() {
         Assert.assertTrue("Row 1 was still selected after de-selecting it.",
                 !grid.getRow(1).isSelected());
     }
+
+    @Test
+    public void gridSetItemsAndSelectItemAndSetItemAfterSetDeselectAllowedFalse() {
+        open();
+
+        // De-selection is allowed(deselectAllowed is false) and then setting items for grid
```
tomivirkki: Grid de-selection is not allowed at first (it's disallowed initially at `setDeselectAllowedAndSetItems`)

vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/GridSingleSelectionIT.java
```diff
@@ -85,4 +85,25 @@ public void checkDeselectionDisallowedInitially() {
...
+        GridElement grid = $(GridElement.class)
+                .id(GridSingleSelectionPage.ITEMS_GRID);
+
+        // Set Items again by clicking the button
+        $("button").id(GridSingleSelectionPage.SET_ITEMS)
```
tomivirkki: The button seems to do a lot more than just setting the items = it also calls `removeAllColumns`, `addColumn`, `setHeightByRows`, `setId` and also re-adds the whole grid (so it ends up being located after the button)

vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/AbstractGridSingleSelectionModel.java
```diff
@@ -83,6 +92,11 @@ public void select(T item) {
         }
     }
 
+    private void implementSelect(T item) {
```
tomivirkki: `implementSelect` doesn't sound right

vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/AbstractGridSingleSelectionModel.java
```diff
@@ -72,9 +72,18 @@ public void select(T item) {
             return;
         }
         T oldItem = selectedItem;
-        doSelect(item, false);
 
-        getGrid().doClientSideSelection(Collections.singleton(item));
+        // Check if it is in case setItems
```
tomivirkki: This just mentions one use-case which causes an issue but it's doesn't say anything about why the following code is needed in case `setItems` is used...

vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/GridSingleSelectionIT.java
hongduc-phan: correct cmt

vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/GridSingleSelectionIT.java
hongduc-phan: separated the method

vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/AbstractGridSingleSelectionModel.java
hongduc-phan: removed the method.

vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/AbstractGridSingleSelectionModel.java
hongduc-phan: changed to fix the bug in gridConnector.js

vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/GridSingleSelectionIT.java
```diff
@@ -85,4 +85,25 @@ public void checkDeselectionDisallowedInitially() {
         Assert.assertTrue("Row 1 was still selected after de-selecting it.",
                 !grid.getRow(1).isSelected());
     }
+
+    @Test
+    public void gridSetItemsAndSelectItemAndSetItemAfterSetDeselectAllowedFalse() {
```
tomivirkki: This test case doesn't seem to be valid. The build passes fine without the changes to the connector.

vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
```diff
@@ -165,20 +165,23 @@
...
-          }
-          const isSelectedItemDifferentOrNull = !grid.activeItem || !item || item.key != grid.activeItem.key;
-          if (!userOriginated && selectionMode === 'SINGLE' && isSelectedItemDifferentOrNull) {
-            grid.activeItem = item;
-            grid.$connector.activeItem = item;
-          }
-        });
+            const isSelectedItemDifferentOrNull = !grid.activeItem || !item || item.key != grid.activeItem.key;
+            if (!userOriginated && selectionMode === 'SINGLE' && isSelectedItemDifferentOrNull) {
+              grid.activeItem = item;
```
tomivirkki: Now that this isn't run for `null` value, `grid.activeItem` has an invalid value after deselection. It should be `null` but it still has an object reference. The fix should probably be done in `__activeItemChanged`

vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/GridSingleSelectionIT.java
hongduc-phan: after discussed, I corrected the test and it covers the bug

vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
hongduc-phan: have a new commit to fix the bug in __activeItemChanged.
Explain the fix: after investigating the bug, the bug comes from when setItems after select(items), the key of select items still there, and original code from master, switch the oldVal and newVal in if (!grid.$connector.deselectAllowed && oldVal) { ..... } , the condition is triggered for setItems, in observer __activeItemChanged. With this switch, it will go else if (!selectedKeys[newVal.key]) { grid.$connector.doSelection([newVal], true); } => key is unknown => bug.
My idea is if the case is raised , I call doDeselection with item is null and it will do nothing in doDeselection method.

tomivirkki: Why would you want to call `doDeselection` with `[null]`if it does nothing??? The result is the same as removing the line altogether.

Also, now that you removed `grid.activeItem = oldVal;`, active item gets set to `null` while there still is a selected item (`activeItem` and `selectedItems` are no longer in sync).

hongduc-phan:
My fix has some issue, it does not cover all cases of the bug. So I continue to implement it and close the PR.

vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
hongduc-phan: new fix. PTAL it.

vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
```diff
@@ -170,14 +172,17 @@
             selectedKeys[item.key] = item;
             if (userOriginated) {
               item.selected = true;
-              grid.$server.select(item.key);
+              if (isNotSwitchNewAndOldValue) {
```
tomivirkki: This logic doesn't work! This flag is set to `false` when deselection is disabled and the user tries to deselect. The following selection of another item would not get communicated to the server because of the global falsy flag. Add a test for this case.

Rather fix this by allowing `activeItem` to be set `null` programmatically (by `doDeselection`).

hongduc-phan: as discussion in person, I pushed the new fix with adding more test that covers the case: a selection of another item would get communicated to the server.

vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
```diff
@@ -194,7 +194,7 @@ import { ItemCache } from '@vaadin/vaadin-grid/src/vaadin-grid-data-provider-mix
           return;
         }
         if (!newVal) {
-          if (!grid.$connector.deselectAllowed && oldVal) {
+          if (!grid.$connector.deselectAllowed && oldVal && selectedKeys[oldVal.key]) {
```
tomivirkki: The logic flow could use some cleanup:
```
if (a && b && c) {
  foo();
} else if (b && c) {
  bar();
}
```
->

```
if (b && c) {
  if (a) {
    foo();
  } else {
    bar();
  }
}
```

hongduc-phan: done.

PR #863, opened 2020-01-15
Remove usage of DndUtil which is in Flow internal

mehdi-vaadin:
Since `com.vaadin.flow.component.dnd.internal` is not among exported packages of `flow-dnd` in an OSGi environment, we shouldn't use `DndUtil` class which is in that package. So instead of calling `DndUtil.addMobileDndPolyfillIfNeeded` directly, `DropTarget.create` and `DragSource.create` are called.

See https://github.com/vaadin/flow/pull/7291 for more details.

### DIFF
--- a/vaadin-grid-flow/src/test/java/com/vaadin/flow/component/grid/GridDndMobilePolyfillTest.java
+++ b/vaadin-grid-flow/src/test/java/com/vaadin/flow/component/grid/GridDndMobilePolyfillTest.java
@@ -1,0 +1,84 @@
+/*
+ * Copyright 2000-2020 Vaadin Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.vaadin.flow.component.grid;
+
+import java.util.Properties;
+
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.Mockito;
+
+import com.vaadin.flow.component.UI;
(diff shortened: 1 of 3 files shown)


mehdi-vaadin:
Fix https://github.com/vaadin/flow/issues/7288

vaadin-grid-flow/src/test/java/com/vaadin/flow/component/grid/GridDndMobilePolyfillTest.java
```diff
@@ -0,0 +1,84 @@
...
+        };
+        ui.getInternals().setSession(session);
+
+        grid = new Grid<>();
+        ui.add(grid);
+    }
+
+    @Test
+    public void gridDnd_setDropMode_mobilePolyfillShouldBeAdded() {
+        grid.setDropMode(GridDropMode.ON_GRID);
```
pleku: Since there is no guarantee that the setup phase has not added already a pending JS invocation, you should do a dump before calling the grid method to make sure the added invocation is not from somewhere else.

vaadin-grid-flow/src/test/java/com/vaadin/flow/component/grid/GridDndMobilePolyfillTest.java
```diff
@@ -0,0 +1,84 @@
...
+        Assert.assertEquals(
+                "Grid::setDropMode should add DnD mobile polyfill script to the UI.",
+                1, ui.getInternals().dumpPendingJavaScriptInvocations().size());
+    }
+
+    @Test
+    public void gridDnd_setRowsDraggable_mobilePolyfillShouldBeAdded() {
+        grid.setRowsDraggable(true);
```
pleku: Same: Since there is no guarantee that the setup phase has not added already a pending JS invocation, you should do a dump before calling the grid method to make sure the added invocation is not from somewhere else.

vaadin-grid-flow/src/test/java/com/vaadin/flow/component/grid/GridDndMobilePolyfillTest.java
mehdi-vaadin: Done

vaadin-grid-flow/src/test/java/com/vaadin/flow/component/grid/GridDndMobilePolyfillTest.java
mehdi-vaadin: Done

PR #868, opened 2020-01-23
Width of Selection column should adapt to the width of checkbox.

### LINKED ISSUE #660: Selection column width should adapt to width of checkbox
Currently the `vaadin-grid-flow-selection-column` has a hardcoded width of `58px`:

https://github.com/vaadin/vaadin-grid-flow/blob/9d9406293526dd0e497b49f5b6b4186e9c6a3d2f/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/vaadin-grid-flow-selection-column.html#L37-L40

This can be too wide and I couldn't find a way to override this. Now that `vaadin-grid-column` has the `autoWidth` property, it would be great to be the default and also let the developer get an instance of the selection column and set the desired width.

hongduc-phan:
Fixes #660

### DIFF
--- a/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/vaadin-grid-flow-selection-column.js
+++ b/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/vaadin-grid-flow-selection-column.js
@@ -36,7 +36,7 @@ import { GridColumnElement } from '@vaadin/vaadin-grid/src/vaadin-grid-column.js
          */
         width: {
           type: String,
-          value: '58px'
+          value: '56px'
         },
 
         /**
--- a/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/vaadin-grid-flow-selection-column.js
+++ b/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/vaadin-grid-flow-selection-column.js
@@ -24,7 +24,7 @@ import { GridColumnElement } from '@vaadin/vaadin-grid/src/vaadin-grid-column.js
       return {
 
         /**
-         * set auto-width is true
+         * Automatically sets the width of the column based on the column contents when this is set to `true`.
          */
         autoWidth: {
           type: Boolean,
--- a/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/vaadin-grid-flow-selection-column.js
+++ b/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/vaadin-grid-flow-selection-column.js
@@ -24,6 +24,14 @@ import { GridColumnElement } from '@vaadin/vaadin-grid/src/vaadin-grid-column.js
       return {
 
         /**
+         * set auto-width is true
+         */
+        autoWidth: {
+          type: Boolean,
+          value: true
+        },
+
+        /**
          * Width of the cells for this column.
          */
         width: {
--- a/vaadin-grid-flow-demo/src/main/java/com/vaadin/flow/component/grid/demo/GridDemo.java
+++ b/vaadin-grid-flow-demo/src/main/java/com/vaadin/flow/component/grid/demo/GridDemo.java
@@ -871,13 +871,6 @@ public class GridDemo extends DemoView {
         grid.addColumn(Person::getAge).setHeader("Age");
 
         grid.setSelectionMode(SelectionMode.MULTI);
-        ((GridMultiSelectionModel<?>) grid
-                .setSelectionMode(Grid.SelectionMode.MULTI))
-                .setAutoWidthSelectionColumn(true);
-        ((GridMultiSelectionModel<?>) grid
-                .setSelectionMode(Grid.SelectionMode.MULTI))
-                .setWidthSelectionColumn("150px");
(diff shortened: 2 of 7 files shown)


vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/vaadin-grid-flow-selection-column.js
```diff
@@ -24,11 +24,11 @@ import { GridColumnElement } from '@vaadin/vaadin-grid/src/vaadin-grid-column.js
       return {
 
         /**
-         * Width of the cells for this column.
+         * Set autoWidth is true.
          */
-        width: {
-          type: String,
-          value: '58px'
```
tomivirkki: Should we still keep this `58px` value as the default? In 99% of the cases it's still a better guess than the `100px` value derived from `vaadin-grid-column` otherwise. The jump that occurs on auto-width might be smaller with this value.

hongduc-phan: yes, it should not replace harded-value width (58px) with auto-width as the default.
But I think the width of flow selection column should adapt to width of checkbox.
So I have new 2 methods: setWidth and setAutoWidth for the flow-selection-column, and we can set width and auto-width.

tomivirkki: No new methods (features) in a bugfix. A user doesn't have access to the selection column and thus can't set an explicit width for it so we're not breaking any behavior by making auto-width the default here, right?

vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/vaadin-grid-flow-selection-column.js
```diff
@@ -23,6 +23,14 @@ import { GridColumnElement } from '@vaadin/vaadin-grid/src/vaadin-grid-column.js
     static get properties() {
       return {
 
+        /**
+         * set auto-wdith is true
+         */
+        autoWidth: {
+          type: String,
```
tomivirkki: Type should be `Boolean`

hongduc-phan: sorry, my bad

vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/vaadin-grid-flow-selection-column.js
hongduc-phan: yes, got it. no feature for fixing bug and no breaking any behavior by adding new features.  I understand it after asking.

vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/vaadin-grid-flow-selection-column.js
```diff
@@ -23,6 +23,14 @@ import { GridColumnElement } from '@vaadin/vaadin-grid/src/vaadin-grid-column.js
     static get properties() {
       return {
 
+        /**
+         * set auto-width is true
```
tomivirkki: This should be the description of the property, not about setting it to some value. Maybe just copy the first line from https://github.com/vaadin/vaadin-grid/blob/master/src/vaadin-grid-column.html#L548 instead (no need to copy the whole description as it's quite verbose)

hongduc-phan: yes, I corrected it to "Automatically sets the width of the column based on the column checkbox when this is set to `true`."

hongduc-phan: corrected it to ...column content...

vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/vaadin-grid-flow-selection-column.js
```diff
@@ -23,6 +23,14 @@ import { GridColumnElement } from '@vaadin/vaadin-grid/src/vaadin-grid-column.js
...
       return {
 
+        /**
+         * Automatically sets the width of the column based on the column contents when this is set to `true`.
+         */
+        autoWidth: {
+          type: Boolean,
+          value: true
+        },
+
```
tomivirkki: The change seems to work fine 👍  One more fine-tune: Once the initial data is received, it looks like the selection column snaps from the default 58px to 56px (by auto-width). So let's align the default width to be `56px` also to avoid the jump in most cases.

hongduc-phan: done with the newest commit.

hongduc-phan: done with the newest commit.

PR #870, opened 2020-01-23
Cherry-pick/autoWidth of flow-selection-column

### LINKED ISSUE #660: Selection column width should adapt to width of checkbox

hongduc-phan:
Fixes #660


hongduc-phan:
I created this PR ti cherry-pick merge to 4.1. But I should close it because it should be created after reviewing code.

PR #873, opened 2020-02-05
Fixes: Reattaching a Grid resets its sort order #747

### LINKED ISSUE #747: Bug: Reattaching a Grid resets its sort order
**Vaadin Version**
14.0.2

**Problem**
If you have a Grid with a surrounding layout and you reattach the Grid it loses its sort order.
To be more precise it overrwrites it with an empty one.

The example code below logs changes to the sort order.

**Example**
As an example i pressed the first, second, third and second columns header.
The log result looks like this:

    1
    2
    3
    3

image

If you press the "rm" button and then the "attach" button `0` gets added to the log and after a short moment the grid loses its sort order.

**Example code**

    @Route("test")
    public class TEST extends FlexLayout {
        private TEST_GUI gui;
    
        public TEST() {
            setSizeFull();
        
            gui = new TEST_GUI();
        
            Button btRm = new Button("rm", evt -> remove(gui));
        
            Button btattach = new Button("attach", evt -> add(gui));
        
            add(btRm, btattach, gui);
        }
    
        private List<TEST_DATA> createTestData() {
            List<TEST_DATA> data = new ArrayList<>();
        
            for(int i=0; i<999; i++) {
                data.add(new TEST_DATA("test" + i, i, (int)(Math.random() * 100)));
            }
        
            return data;
        }
    
        public class TEST_GUI extends FlexLayout {
            public TEST_GUI() {
                setSizeFull();
            
                Grid<TEST_DATA> grid = new Grid<>(TEST_DATA.class);
                grid.setSizeFull();
            
                grid.setItems(createTestData());
                // removes generated columns
                grid.removeAllColumns();
            
                grid.addColumn(TEST_DATA::getStest).setSortable(true).setHeader("String");
                grid.addColumn(TEST_DATA::getSint).setSortable(true).setHeader("index");
                grid.addColumn(TEST_DATA::getRandom).setSortable(true).setHeader("random");
            
                grid.setMultiSort(true);
                grid.setSelectionMode(SelectionMode.MULTI);
            
                grid.addSortListener(e -> {
                    System.out.println(e.getSortOrder().size());
                });
                
                add(grid);
            }
        }
    
        public class TEST_DATA {
            public String stest;
            public int sint;
            public int random;
        
            public TEST_DATA(String stest, int sint, int random) {
                super();
                this.stest = stest;
                this.sint = sint;
                this.random = random;
            }
            public String getStest() {
                return stest;
            }
            public void setStest(String stest) {
                this.stest = stest;
            }
            public int getSint() {
                return sint;
            }
            public void setSint(int sint) {
                this.sint = sint;
            }
            public int getRandom() {
                return random;
            }
            public void setRandom(int random) {
                this.random = random;
            }
        }
    }

hongduc-phan:
The fix is to address the issue: deattach and reattach/refresh or something, the grid will keep the sort state.
Fixes #747


vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/it/PreserveOnRefreshPage.java
```diff
@@ -32,6 +39,96 @@ public PreserveOnRefreshPage() {
         grid.addComponentColumn(person -> new Span(person.getFirstName()))
                 .setHeader(new Span("header")).setFooter(new Span("footer"));
         add(grid);
+        TestGui();
```
tomivirkki: These tests don't belong in the PreserveonRefreshPage. This is related to sorting so that's where it should be. SortingPage also already has a grid with sorting so you should only add the detach/attach buttons there.

hongduc-phan:
> Why does the new test case pass? The changes in this PR at least don't fix the original bug. This indicates the test case is also invalid.

Moved the test and will update the test with assert sort state

hongduc-phan:
> Why does the new test case pass? The changes in this PR at least don't fix the original bug. This indicates the test case is also invalid.

My current test is to cover checking the order of items after re-attaching. I will add more test to cover the sorting state

vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/it/PreserveOnRefreshPage.java
hongduc-phan: Thank you for correcting me. moved the test to the right place. and my PR is already review.

vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
```diff
@@ -398,15 +398,52 @@ import { ItemCache } from '@vaadin/vaadin-grid/src/vaadin-grid-data-provider-mix
       })
 
       const sorterChangeListener = tryCatchWrapper(function(_, oldValue) {
+        let isSortersDuplicate = false;
+        let sortersDuplicate = [];
+        console.log(grid._sorters)
```
tomivirkki: No console logs

vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
```diff
@@ -398,15 +398,52 @@ import { ItemCache } from '@vaadin/vaadin-grid/src/vaadin-grid-data-provider-mix
...
+                break;
+              }
+            }
+            if(isSortersDuplicate) {
+              break;
+            }
+          }
+        }
+        if(sortersDuplicate.length > 0 ) {
+          let tempGrid = grid._sorters;
```
tomivirkki: `tempGrid` as a variable name for for `grid._sorters`?

vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
```diff
@@ -3004,11 +3013,39 @@ private void setSortOrder(List<GridSortOrder<T>> order,
...
+
+        boolean isSortOrderNotEmptyAfterReAttached = false;
+        if (order.isEmpty() && isSorted && !sortOrder.isEmpty() && !isPreviousSortValueNotEmpty) {
+            isSortOrderNotEmptyAfterReAttached = true;
+        }
+        // In case that reattach the grid or something to keep the sort-state
+        // exception cases: reset, or set back to sort size = 0
+        if (!isResetSort && isSortOrderNotEmptyAfterReAttached) {
+            order = sortOrder.stream().collect(Collectors.toList());
+            updateClientSideSorterIndicators(order);
```
tomivirkki: `updateClientSideSorterIndicators` is now invoked in 4 different places with different conditions within the same method! Does that 1 added test case really cover all of these cases? You should rethink your approach to fixing this issue as this is getting really complex.

vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
hongduc-phan: removed.

vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
hongduc-phan: changed name to tempSorters to make more sense.

vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
```diff
@@ -3004,11 +3013,39 @@ private void setSortOrder(List<GridSortOrder<T>> order,
...
 
+        boolean isOldValSameNewValWithoutOrder = false;
+        List<GridSortOrder<T>> tempSortOrder = sortOrder.stream().collect(Collectors.toList());
+        if (tempSortOrder.containsAll(order) && order.containsAll(tempSortOrder)) {
+            isOldValSameNewValWithoutOrder = true;
+        }
+
         if (!userOriginated) {
             updateClientSideSorterIndicators(order);
+        } else if (userOriginated
```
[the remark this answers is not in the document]
hongduc-phan: removed it. it is my bad. when coding, I was thinking about another case that is`userOriginated`, and I type it to code.

tomivirkki:
This fix is getting a bit too complex. The original issue is that the `sortOrder` state of the server-side Grid is not synchronized to the client-side when the component is reattached. Fixing that should be enough: add an attach listener to the Grid and invoke the existing method that synchronizes the `sortOrder` state to the client inside the listener.

vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
```diff
@@ -2999,6 +3000,14 @@ public void sort(List<GridSortOrder<T>> order) {
         setSortOrder(order, false);
     }
 
+    @Override
+    protected void onAttach(AttachEvent attachEvent) {
+        super.onAttach(attachEvent);
+        if (!sortOrder.isEmpty()) {
```
tomivirkki: Is this check necessary? If it's only for optimization, it's probably not worth the extra lines. Logically `sortOrder` should be synchronized to the client even if it's emptied; the fact that a new Web Component instance is created on attach is an implementation detail and it's preferable for code not to rely on that.

hongduc-phan: removed it

vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
```diff
@@ -398,14 +398,12 @@ import { ItemCache } from '@vaadin/vaadin-grid/src/vaadin-grid-data-provider-mix
...
-              direction: sorter.direction
-            };
-          }));
-        }
+        grid.$server.sortersChanged(grid._sorters.map(function(sorter) {
+          return {
+            path: sorter.path,
+            direction: sorter.direction
+          };
+        }));
```
tomivirkki: Why is this change needed? Is it tested?

vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/it/SortingPage.java
```diff
@@ -52,6 +57,80 @@ public SortingPage() {
...
+
+    public class TestGui extends FlexLayout {
+        Grid<TestData> grid = new Grid<>(TestData.class);
+        public TestGui() {
+            setSizeFull();
+            grid.setSizeFull();
+            grid.setItems(createTestData());
+            grid.setId("second-grid");
+            // removes generated columns
+            grid.removeAllColumns();
```
tomivirkki: Don't create the grid with the bean class constructor if you want to define the columns manually.

vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/it/SortingPage.java
```diff
@@ -52,6 +57,80 @@ public SortingPage() {
...
+                data.add(new TestData("test" + i, i, 3 * 100));
+            }
+            return data;
+        }
+    }
+
+    public class TestData {
+        private String stest;
+        private int sint;
+        private int random;
```
tomivirkki: The naming of "stest", "sint", "random" is unclear.

vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/it/SortingPage.java
```diff
@@ -52,6 +57,80 @@ public SortingPage() {
...
+        btnReset.setId("btn-reset");
+        Button btRm = new Button("deattach", evt -> remove(gui1));
+        btRm.setId("btn-deattach");
+        Button btattach = new Button("attach",
+                evt -> add(gui1));
+        btattach.setId("btn-attach");
+        add(btRm, btattach,btnReset);
+    }
+
+    public class TestGui extends FlexLayout {
```
tomivirkki: Why can't we use the existing grid and data for the new test cases?

vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/it/SortingPage.java
```diff
@@ -52,6 +57,80 @@ public SortingPage() {
...
+
+            add(grid);
+        }
+        public Grid<TestData> getGrid() {
+            return this.grid;
+        }
+        private List<TestData> createTestData() {
+            List<TestData> data = new ArrayList<>();
+            for(int i=0; i<9; i++) {
+                data.add(new TestData("test" + i, i, 3 * 100));
```
tomivirkki: 3 * 300 isn't really "random". Random data should be avoided in tests anyway.

vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/SortingIT.java
```diff
@@ -61,6 +62,141 @@ public void setInitialSortOrder_changeOrderFromServer_sortIndicatorsUpdated() {
...
+        String textFirstStringColumnAfterReattch = secondGridElement.getCell(0, 1).getText();
+        String textFirstIndexColumnAfterReattch = secondGridElement.getCell(0, 2).getText();
+        String textFirstRandomColumnAfterReattch = secondGridElement.getCell(0, 3).getText();
+        Assert.assertEquals("asc", secondGrid.findElements(By.tagName("vaadin-grid-sorter"))
+                .get(1).getAttribute("direction"));
+        Assert.assertEquals(textFirstStringColumnBeforeReattch, textFirstStringColumnAfterReattch);
+        Assert.assertEquals(textFirstIndexColumnBeforeReattch, textFirstIndexColumnAfterReattch);
+        Assert.assertEquals(textFirstRandomColumnBeforeReattch, textFirstRandomColumnAfterReattch);
+
+        // Assert sort with rather than 1 column, handle from client
```
tomivirkki: This is a bit hard to understand. Can you rephrase?

vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/it/SortingPage.java
```diff
@@ -52,6 +57,80 @@ public SortingPage() {
...
+            grid.setSizeFull();
+            grid.setItems(createTestData());
+            grid.setId("second-grid");
+            // removes generated columns
+            grid.removeAllColumns();
+            Grid.Column<TestData> string = grid.addColumn(TestData::getStest).setSortable(true).setHeader("String");
+            Grid.Column<TestData> index = grid.addColumn(TestData::getSint).setSortable(true).setHeader("index");
+            grid.addColumn(TestData::getRandom).setSortable(true).setHeader("random");
+            grid.setMultiSort(true);
+            grid.setSelectionMode(Grid.SelectionMode.MULTI);
```
tomivirkki: Why is selection mode needed in the tests?

vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/it/SortingPage.java
```diff
@@ -15,14 +15,17 @@
  */
 package com.vaadin.flow.component.grid.it;
 
+import java.util.ArrayList;
 import java.util.List;
 
+import com.vaadin.flow.component.button.Button;
```
tomivirkki: Now the test UI has both NativeButtons and Buttons.

vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/it/SortingPage.java
```diff
@@ -52,6 +57,80 @@ public SortingPage() {
         });
         button.setId("sort-by-age");
         add(button);
+
+        setSizeFull();
```
tomivirkki: No need to use setSizeFull for this, the gui and the grid.

vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/it/SortingPage.java
hongduc-phan: in the first time, I used the existing grid, but after fixed the bug (reattach and grid keeps the sort state, there is a new issue that the number of vaadin-grid-sorter is increased after detach/reattach-ing grid. and it is quite hard to assert if the grid has only 2 columns and we just have sort number state is 1-2 or 2-1. That is why I have to add a new grid to check the number like 1-2-3 or 1-3-2 or 3-1-2. But this morning, we agree that is another issue and we will separate it. The new issue is that if clicking re(attach) is more 2 times without detaching, the grid can not keep the sort state.

So I will back to work on the existing grid because I just check the grid keeps the sort state after the grid is reattached or refresh the page.

tomivirkki: If a third column is needed (not sure if it is), you can add a new column to the existing grid.

hongduc-phan: I backed to work on the existing grid, edited tests with the existing grid.
And with this issue, we do not need the second grid.
in the future, when working on the issue that reattaches much time without detaching, maybe we can add new column for the test.

vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/SortingIT.java
hongduc-phan: it means that I click and sort column, from client, not using sort(abc) from java side. 
And sort with 2-3-4 columns, instead only one column.

hongduc-phan: but it is an old test, my new test to test the flow: attach - detach and attach again, and check the sort states are kept or not.

vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/SortingIT.java
```diff
@@ -61,6 +62,57 @@ public void setInitialSortOrder_changeOrderFromServer_sortIndicatorsUpdated() {
...
         assertAscendingSorter("Age");
     }
+    @Test
+    public void keepSortStatesAfterReAttach() {
+        WebElement btnAttach = findElement(By.id("btn-attach"));
+        WebElement btnRemove = findElement(By.id("btn-detach"));
+        GridElement sortingGridElement = $(GridElement.class).first();
+        findElement(By.id("sort-by-age")).click();
+        // Attach grid
+        btnAttach.click();
```
tomivirkki: The grid is already attached

vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/SortingIT.java
```diff
@@ -61,6 +62,57 @@ public void setInitialSortOrder_changeOrderFromServer_sortIndicatorsUpdated() {
...
+        WebElement sortingGrid = findElement(By.id("sorting-grid"));
+        Assert.assertEquals("asc", sortingGrid.findElements(By.tagName("vaadin-grid-sorter"))
+                .get(1).getAttribute("direction"));
+        String textAgeColumnBeforeReattch = sortingGridElement.getCell(0, 1).getText();
+        // Detach
+        btnRemove.click();
+        // Reattach
+        btnAttach.click();
+        sortingGridElement = $(GridElement.class).first();
+        sortingGrid = findElement(By.id("sorting-grid"));
```
tomivirkki: Isn't `sortingGridElement` and `sortingGrid` the same exact object?

vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/SortingIT.java
```diff
@@ -61,6 +62,57 @@ public void setInitialSortOrder_changeOrderFromServer_sortIndicatorsUpdated() {
...
+        Assert.assertEquals(textAgeColumnBeforeReattch, textAgeColumnAfterReattch);
+    }
+
+    @Test
+    public void indicatorsSortStateNumbers() {
+        WebElement btnAttach = findElement(By.id("btn-attach"));
+        WebElement btnRemove = findElement(By.id("btn-detach"));
+
+        // Attach grid
+        btnAttach.click();
```
tomivirkki: The grid is already attached

vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/SortingIT.java
```diff
@@ -61,6 +62,57 @@ public void setInitialSortOrder_changeOrderFromServer_sortIndicatorsUpdated() {
...
+        findElement(By.id("sort-by-age")).click();
+        WebElement sortingGrid = findElement(By.id("sorting-grid"));
+        Assert.assertEquals("asc", sortingGrid.findElements(By.tagName("vaadin-grid-sorter"))
+                .get(1).getAttribute("direction"));
+
+        String sortStateNumberAgeColumn
+                = sortingGrid.findElements(By.tagName("vaadin-grid-sorter")).get(1).getAttribute("_order");
```
tomivirkki: Sort order isn't used/tested in this test case. Just one column is sorted so `_order` is always `null`.

vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/SortingIT.java
```diff
@@ -61,6 +62,57 @@ public void setInitialSortOrder_changeOrderFromServer_sortIndicatorsUpdated() {
...
+        // Detach
+        btnRemove.click();
+        // Reattach
+        btnAttach.click();
+        sortingGrid = findElement(By.id("sorting-grid"));
+        Assert.assertEquals("asc", sortingGrid.findElements(By.tagName("vaadin-grid-sorter"))
```
tomivirkki: Didn't you already test this in the previous test case?

vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/SortingIT.java
hongduc-phan: removed the line attach grid.

vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/SortingIT.java
hongduc-phan: removed the line attach grid.

vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/SortingIT.java
hongduc-phan: I refactored code and combined 2 tests into one. the test checks content of first row, number sort state and direction changes after reattach or not.

vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/SortingIT.java
hongduc-phan: I changed the test, and the grid is now multiSort. I check 2 columns are sorted

vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/SortingIT.java
hongduc-phan: The first one, I want to take a row, cell and so on of grid (GridElement class)
The second, I want to take the tag name of vaadin-grid-sorter, from the grid, to limit the range when using findElements. but I removed the second because I think it not necessary. there is only one grid.

tomivirkki: Yes, the GridElement is also a WebElement

PR #874, opened 2020-02-06
Bump WC version and add a test

### LINKED ISSUE #855: recalculateColumnWidth() does not work within ExpandListener
When expanding a node with a subtree, the expand listener is called before the client-side
is finished with expanding.
If you call recalculateWidth() within the expand listener code, then it does not have an effect for us.

However, if you create a button and a click listener and put the recalculateWidth() call there, it works fine -- the obvious difference is that the call is executed *after* the expansion is done in the browser.

hongduc-phan:
Fixes in WC but the original bug in Flow-component. So the test is to reproduce the bug in Flow-Component.

Fixes #855

### DIFF
--- a/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/treegrid/it/TreeGridExpandAllIT.java
+++ b/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/treegrid/it/TreeGridExpandAllIT.java
@@ -17,6 +17,8 @@ package com.vaadin.flow.component.treegrid.it;
 
 import java.util.List;
 
+import com.vaadin.flow.component.grid.testbench.GridTHTDElement;
+import com.vaadin.flow.component.grid.testbench.TreeGridElement;
 import org.junit.Assert;
 import org.junit.Test;
 import org.openqa.selenium.By;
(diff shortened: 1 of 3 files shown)


vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/treegrid/it/TreeGridExpandAllIT.java
```diff
@@ -46,6 +50,20 @@ public void addNewItemAfterCollapseAndExpand() throws InterruptedException {
         assertNewChild();
     }
 
+    @Test
+    public void recalculateWidthsAfterExpend() throws InterruptedException {
+        open();
+        TreeGridElement grid = $(TreeGridElement.class).get(1);
+        WebElement expandToggleElement = grid.getExpandToggleElement(1, 0);
+        int widthBeforeExpend = grid.findElements(By.tagName("vaadin-grid-cell-content")).get(2).getSize().getWidth();
```
tomivirkki: "2" here is a magic number. Rather use something else, like `vaadin-grid-tree-toggle` element if it works for this case. If it doesn't, you should add a dedicated element for this measuring purpose.

hongduc-phan: thank you for your correct. changed it.

PR #916, opened 2020-03-25
Add drag and drop demo for tree grid

shadikhani:
1- I didn't use the sample code. I created a simple drag and drop example.
2- This demo has a problem that seems to be a bug in Flow. It does not work when dragging the last item of a parent. I will create an issue later.

Fix https://github.com/vaadin/90min-dx-tasks/issues/103

### DIFF
--- a/vaadin-grid-flow-demo/src/main/java/com/vaadin/flow/component/treegrid/demo/TreeGridDemo.java
+++ b/vaadin-grid-flow-demo/src/main/java/com/vaadin/flow/component/treegrid/demo/TreeGridDemo.java
@@ -3,15 +3,18 @@ package com.vaadin.flow.component.treegrid.demo;
 import com.vaadin.flow.component.Component;
 import com.vaadin.flow.component.dependency.JsModule;
 import com.vaadin.flow.component.grid.demo.GridDemo.Person;
+import com.vaadin.flow.component.grid.dnd.GridDropMode;
+import com.vaadin.flow.component.html.Div;
+import com.vaadin.flow.component.html.NativeButton;
+import com.vaadin.flow.component.notification.Notification;
+import com.vaadin.flow.component.textfield.TextArea;
+import com.vaadin.flow.component.treegrid.TreeGrid;
 import com.vaadin.flow.component.treegrid.demo.data.DepartmentData;
 import com.vaadin.flow.component.treegrid.demo.entity.Account;
 import com.vaadin.flow.component.treegrid.demo.entity.Department;
 import com.vaadin.flow.component.treegrid.demo.service.AccountService;
-import com.vaadin.flow.component.html.Div;
-import com.vaadin.flow.component.html.NativeButton;
-import com.vaadin.flow.component.textfield.TextArea;
-import com.vaadin.flow.component.treegrid.TreeGrid;
 import com.vaadin.flow.data.provider.hierarchy.AbstractBackEndHierarchicalDataProvider;
+import com.vaadin.flow.data.provider.hierarchy.AbstractHierarchicalDataProvider;
 import com.vaadin.flow.data.provider.hierarchy.HierarchicalDataProvider;
 import com.vaadin.flow.data.provider.hierarchy.HierarchicalQuery;
 import com.vaadin.flow.demo.DemoView;
@@ -44,6 +47,7 @@ public class TreeGridDemo extends DemoView {
     protected void initView() {
         createBasicTreeGridUsage();
         createLazyLoadingTreeGridUsage();
+        createDragAndDrop();
     }
 
     private void createBasicTreeGridUsage() {
@@ -119,8 +123,7 @@ public class TreeGridDemo extends DemoView {
         grid.addHierarchyColumn(Account::toString).setHeader("Account Title");
         grid.addColumn(Account::getCode).setHeader("Code");
 
-        HierarchicalDataProvider dataProvider =
-                new AbstractBackEndHierarchicalDataProvider<Account, Void>() {
+        HierarchicalDataProvider dataProvider = new AbstractBackEndHierarchicalDataProvider<Account, Void>() {
 
             @Override
             public int getChildCount(HierarchicalQuery<Account, Void> query) {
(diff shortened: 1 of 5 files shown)


vaadin-grid-flow-demo/src/main/java/com/vaadin/flow/component/treegrid/demo/TreeGridDemo.java
```diff
@@ -146,4 +149,78 @@ public boolean hasChildren(Account item) {
...
         addCard("TreeGrid with lazy loading", grid);
     }
+
+    // Drag and drop
+    private Department draggedItem;
+
+    private void createDragAndDrop() {
+        DepartmentData departmentData = new DepartmentData();
+        // begin-source-example
+        // source-example-heading: Drag and drop
```
tmattsso: There's no comments on any of this code. At minimum document what part of the code does what.

vaadin-grid-flow-demo/src/main/java/com/vaadin/flow/component/treegrid/demo/TreeGridDemo.java
```diff
@@ -146,4 +149,78 @@ public boolean hasChildren(Account item) {
...
+        TreeGrid<Department> treeGrid = new TreeGrid<>();
+
+        treeGrid.setDataProvider(
+                new AbstractHierarchicalDataProvider<Department, Object>() {
```
tmattsso: Is this code necessary for this demo? Consider abstracting it out to a class or method.

vaadin-grid-flow-demo/src/main/java/com/vaadin/flow/component/treegrid/demo/TreeGridDemo.java
```diff
@@ -146,4 +149,78 @@ public boolean hasChildren(Account item) {
...
+        treeGrid.addHierarchyColumn(Department::getName)
+                .setHeader("Department Name");
+        treeGrid.addColumn(Department::getManager).setHeader("Manager");
+        treeGrid.setRowsDraggable(true);
+        treeGrid.setDropMode(GridDropMode.ON_TOP);
+
+        treeGrid.addDragStartListener(
+                event -> draggedItem = event.getDraggedItems().get(0));
+
+        treeGrid.addDragEndListener(event -> draggedItem = null);
```
tmattsso: draggedItem should be part of the code shown to the user

vaadin-grid-flow-demo/src/main/java/com/vaadin/flow/component/treegrid/demo/TreeGridDemo.java
```diff
@@ -146,4 +149,78 @@ public boolean hasChildren(Account item) {
...
+
+        treeGrid.addDropListener(event -> {
+            Department department = event.getDropTargetItem().get();
```
tmattsso: the variable needs a better name, it wasn't immediately clear what this code was doing.

Is there a way of preventing this drop action completely?  with specifying allowed drop targets or something?

vaadin-grid-flow-demo/src/main/java/com/vaadin/flow/component/treegrid/demo/data/DepartmentData.java
```diff
@@ -8,7 +8,7 @@
 import java.util.stream.Collectors;
 
 public class DepartmentData {
-    private static final List<Department> DEPARTMENT_LIST = createDepartmentList();
+    private final List<Department> DEPARTMENT_LIST = createDepartmentList();
```
tmattsso: Why remove 'static'?

shadikhani: Previously, the grid was read-only, but now it is not. The parent of the dragged items is changing by dragging and dropping items.

vaadin-grid-flow-demo/src/main/java/com/vaadin/flow/component/treegrid/demo/TreeGridDemo.java
shadikhani: Done.

vaadin-grid-flow-demo/src/main/java/com/vaadin/flow/component/treegrid/demo/TreeGridDemo.java
shadikhani: Done.

vaadin-grid-flow-demo/src/main/java/com/vaadin/flow/component/treegrid/demo/TreeGridDemo.java
shadikhani: Done.

vaadin-grid-flow-demo/src/main/java/com/vaadin/flow/component/treegrid/demo/TreeGridDemo.java
shadikhani: Variable name changed.

I have no idea.

tomivirkki:
This PR's base is `master`, which is the V15 version of the component. vcom runs V14 only so the changes of this PR wouldn't end up on the website. We're planning to remove the demo-modules from the `master` branches on some timeframe as they serve no purpose currently. You can change base with the "Edit" button next to the title.

tomivirkki:
The new demo feels buggy.

Kapture 2020-03-31 at 12 38 49

Also, there are D&D demos with TreeGrid already at:

https://github.com/vaadin/vaadin-grid-flow/blob/master/vaadin-grid-flow-demo/src/main/java/com/vaadin/flow/component/grid/demo/GridDemo.java#L2364-L2428

https://github.com/vaadin/vaadin-grid-flow/blob/master/vaadin-grid-flow-demo/src/main/java/com/vaadin/flow/component/grid/demo/GridDemo.java#L2526-L2605

Those could be reused (or copied) in TreeGridDemo.

But still, there are lots of Grid features that aren't demonstrated in the TreeGrid demos (like (multi)selection, sorting, filtering, header/footer config, component renderer, item details, context menu, editor...). Why should D&D be an exception? Wouldn't it make sense that the TreeGrid demos only showcase features specific to TreeGrid? Maybe TreeGrid demos could have a link to Grid demos to emphasize that all Grid features also work with TreeGrid.

tomivirkki:
Closing as inactive.

PR #925, opened 2020-04-03
Bump flow-component-base to 3.1

tulioag:
Broken because of https://github.com/vaadin/flow/pull/7652.


tulioag:
Outdated after #931

PR #932, opened 2020-04-09
Remove FooterRow from Grid (#922)

benno-mueller-saxess:


benno-mueller-saxess:
Sorry for creating a new issue. I'm not so experienced with the GitHub workflow. This should belong to issue #922

PR #950, opened 2020-04-22
Reset TreeGrid connector on refreshAll

### LINKED ISSUE #891: Filtering+expansion broken
Just encountered this issue in Vaadin 14.1.17

The expansion is not only slow, but incorrect as well, at least when driven by valueChangedEvent on a filter field.

If I paste filter-value into my ValueChangeMode.EAGER filter-field, the filtering is correct, but if I type it letter by letter, the finished filtered list is 4x as long as it should be (because my filter had 4 letters?) and it is garbled; repeating sections; empty lines.
If I switch to ValueChangeMode.LAZY it seems to behave better, but I can still get incorrect result.

Next, if I stick with ValueChangeMode.LAZY on filter-field, and include parents in filter, so that when I filter & expand I have some parents with children, but at least one parent without children, the expansion seems to stop. All top-level parent items show with "expanded" icon, but are not. When I "close" one, the others suddenly show their children, but the structure seems broken, and weird stuff happens from then on.

If I only filter on children, then expand all seems to work fine.

_Originally posted by @guttormvik in https://github.com/vaadin/vaadin-grid-flow/issues/476#issuecomment-588962916_

pekam:
Fixes #891.

By clearing the connector from all the pending stuff from previous
refresh, the data is not messed up by updates from outdated requests.

This seems to also help with the slowness of the updates at least to
some extent.

### DIFF
--- a/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/treegrid/it/TreeGridFilterExpandIT.java
+++ b/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/treegrid/it/TreeGridFilterExpandIT.java
@@ -34,7 +34,8 @@ public class TreeGridFilterExpandIT extends AbstractTreeGridIT {
     }
 
     @Test // https://github.com/vaadin/vaadin-grid-flow/issues/891
-    public void filterAndExpandAllMultipleTimes_correctItemsShown() {
+    public void filterAndExpandAllMultipleTimes_correctItemsShown()
+            throws InterruptedException {
         TestBenchElement filterField = $("input").first();
 
         // Typing two letters with ValueChangeMode.EAGER on the field to trigger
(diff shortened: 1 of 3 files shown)


vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/treegrid/it/TreeGridFilterExpandIT.java
```diff
@@ -0,0 +1,71 @@
...
+
+        // Typing two letters with ValueChangeMode.EAGER on the field to trigger
+        // two filtering updates.
+        filterField.sendKeys("ap");
+
+        // Because of the way that the connector forcefully updates grid's
+        // cache, and the incorrect updates are coming asynchronously, there
+        // seems to be no way of checking when all the requests have been
+        // responded, and no more forceful updates are coming.
+        waitForSeconds(2);
```
tulioag, line 48: Would `Thread.sleep(2000)` be simpler?

pekam: Yeah. Done.

PR #968, opened 2020-05-08
Add benchmark tests

tomivirkki:
Relates to https://github.com/vaadin/components-team-tasks/issues/527

Add benchmark tests


vaadin-grid-flow-integration-tests/benchmark/run-benchmark.js
```diff
@@ -0,0 +1,195 @@
...
+    `
+    <artifactId>jetty-maven-plugin</artifactId>
+      <configuration>
+        <httpConnector>
+          <port>${REF_JETTY_PORT}</port>
+        </httpConnector>
+        <stopPort>${REF_JETTY_PORT + 1}</stopPort>
+      </configuration>
+    `
+  );
```
tulioag, line 98: - This seems a bit fragile. Do you think this PR makes sense? https://github.com/vaadin/flow-component-base/pull/179
Then we could just run `mvn jetty:run -Djetty.http.port=SOMEPORT -Djetty.stop.port=SOMEOTHERPORT`

- Should we be more explicit about the stop port? Might make debugging easier in the case jetty fails to run because of port conflicts.

vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/it/GridBenchmark.java
```diff
@@ -0,0 +1,177 @@
...
+import com.vaadin.flow.theme.lumo.Lumo;
+
+@Route("benchmark")
+@Theme(Lumo.class)
+@JsModule("./benchmark.js")
+public class GridBenchmark extends Div implements HasUrlParameter<String> {
+
+    private Grid<String> grid;
+
+    private static List<String> items = IntStream.range(0, 1000).boxed().map(Object::toString)
```
tulioag, line 49: - Nit: Small simplification: `IntStream.range(0,1000).mapToObj(String::valueOf).collect(Collectors.toList())`

- Should this be `final` and wrapped in a UnmodifiableList to avoid unintended modifications?

vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/it/GridBenchmark.java
```diff
@@ -0,0 +1,177 @@
...
+
+        Map<String, List<String>> parametersMap = queryParameters.getParameters();
+        if (!parametersMap.containsKey("variant") || !parametersMap.containsKey("metric")) {
+            add(new Text("Provide query parameters: variant and metric"));
+            return;
+        }
+        String metric = parametersMap.get("metric").get(0);
+        String variant = parametersMap.get("variant").get(0);
+
+        System.out.println("Sample: " + variant + "-" + metric);
```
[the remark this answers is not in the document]
tulioag: Something like this might keep sonarqube happy `LoggerFactory.getLogger(getClass()).info`

vaadin-grid-flow-integration-tests/frontend/benchmark.js
```diff
@@ -0,0 +1,186 @@
...
+
+  // Switch horizontal scroll direction in case end was reached
+  if (deltaXMultiplier) {
+    const overflow = grid.getAttribute('overflow');
+    if (deltaXMultiplier === 1 && !overflow.includes('right')) {
+      deltaXMultiplier = -1;
+    } else if (deltaXMultiplier === -1 && !overflow.includes('left')) {
+      deltaXMultiplier = 1;
+    }
+  }
```
pekam, line 157: This switching of directions didn't work for `multicolumn` variant when I tested the page locally. After reaching the right end, it stays there until the test is finished. With `mixed` variant it worked.

vaadin-grid-flow-integration-tests/frontend/benchmark.js
```diff
@@ -0,0 +1,186 @@
...
+    const listener = (e) => {
+      if (e.animationName === 'content-ready' || e.propertyName === 'opacity') {
+        endTime = performance.now();
+        readyTimer && clearTimeout(readyTimer);
+        if (!grid.loading) {
+          readyTimer = setTimeout(() => {
+            grid.$.scroller.removeEventListener('animationstart', listener);
+            grid.removeEventListener('animationstart', listener);
+            grid.removeEventListener('transitionstart', listener);
+            resolve(performance.now() - endTime);
```
pekam, line 82: Unless I've missed something, this could be simplified by just resolving `endTime`.

Then `measureRender` below would be simplified from:
`reportResult(performance.now() - start - idleTime)`
to:
`reportResult(endTime - start)`

vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/it/GridBenchmark.java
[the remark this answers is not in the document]
tomivirkki: Thx, fixed

vaadin-grid-flow-integration-tests/benchmark/run-benchmark.js
tomivirkki: Yes, would be more solid that way. Then the default 9998 port could also be enforced

vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/it/GridBenchmark.java
tomivirkki: Done.

vaadin-grid-flow-integration-tests/frontend/benchmark.js
```diff
@@ -0,0 +1,186 @@
...
+    };
+
+    grid.$.scroller.addEventListener('animationstart', listener);
+    grid.addEventListener('animationstart', listener);
+    grid.addEventListener('transitionstart', listener);
+  });
+};
+
+const reportResult = (result) => {
+  // @ts-ignore
```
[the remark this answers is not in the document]
tomivirkki: For me, the IDE didn't stop complaining without the markers.

vaadin-grid-flow-integration-tests/frontend/benchmark.js
tomivirkki: That's strange. For me, http://localhost:9998/benchmark?metric=horizontalscrollframetime&variant=multicolumn worked fine on both Chrome and FF

vaadin-grid-flow-integration-tests/frontend/benchmark.js
tomivirkki: Good catch, done.

vaadin-grid-flow-integration-tests/frontend/benchmark.js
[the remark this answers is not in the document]
tomivirkki: Worked around by using `/** @type {any} */`.

vaadin-grid-flow-integration-tests/frontend/benchmark.js
pekam: Hmm... I have tested with Windows Chrome and Linux Mint Chrome and FF, which all reproduce the issue.

Once the scrollbar is within ~15px from the end, events with any positive `deltaX` are dismissed.

@tulioag does this work in your environment?

tomivirkki: Was able to reproduce this by enabling the physical scrollbars. Fixed by having 100px thresholds on both ends instead.

pekam: 👍 seems to work

vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/it/GridBenchmark.java
```diff
@@ -0,0 +1,179 @@
...
+                treeData.getRootItems().forEach(item -> grid.setDetailsVisible(item, true));
+                break;
+            default:
+                break;
+        }
+
+        switch (metric) {
+            case "verticalscrollframetime":
+                add(grid);
+                whenRendered(grid).then(v -> grid.getElement().executeJs("window.measureScrollFrameTime(this, false)"));
```
pekam, line 124: I'm not sure I understand the use of `PendingJavaScriptResult::then` here and other cases below.

You've used it for JavaScript methods that return `Promises`, but I don't think the `PendingJavaScriptResult::then` waits for promises to resolve. Based on the JavaDoc, the callback is simply invoked with the js-function's return value after the roundtrip.

Have I misunderstood the logic here?

vaadin-grid-flow-integration-tests/frontend/benchmark.js
```diff
@@ -0,0 +1,188 @@
...
+    customElements.whenDefined('vaadin-grid'),
+    customElements.whenDefined('vaadin-grid-column'),
+  ]).then(() => (start = performance.now()));
+};
+
+_window.startWhenRendered = (grid) => {
+  return _window.whenRendered(grid).then(() => {
+    start = performance.now();
+  });
+};
```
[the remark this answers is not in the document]
tomivirkki: Good catch. Fixed, it's supposed to be used with the expandtime tests.

vaadin-grid-flow-integration-tests/frontend/benchmark.js
```diff
@@ -0,0 +1,188 @@
...
+
+_window.measureRender = (grid) => {
+  _window.whenRendered(grid).then((endTime) => reportResult(endTime - start));
+};
+
+const SCROLL_TIME = 4000;
+const WARMUP_TIME = 1000;
```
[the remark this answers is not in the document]
tomivirkki: Removed, not worth the extra logic. Initially, the intention was to eliminate the spike before the virtual scrolling kicks in (rows are being translated).

vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/it/GridBenchmark.java
tomivirkki: Seems to properly wait for the promise to resolve. Just tested with
```js
_window.startWhenRendered = (grid) => {
  return new Promise(resolve => {
    setTimeout(() => resolve(), 10000);
  });
};
```

And expandtime tests wait for 10 seconds before starting

pekam: Sorry, my bad. I don't know what docs I was looking at previously, but now I can see the JavaDocs clearly mention that it can handle Promises.

vaadin-grid-flow-integration-tests/benchmark/run-benchmark.js
```diff
@@ -0,0 +1,195 @@
...
+const refGridPath = path.resolve(REF_GRID_DIR);
+
+const gridTestPath = path.resolve(gridPath, TEST_DIR);
+const refGridTestPath = path.resolve(refGridPath, TEST_DIR);
+
+const resultsPath = path.resolve('./results');
+
+const REF_JETTY_PORT = 8088;
+// TODO: Fix as a Grid tag of a revision that will be used as the baseline/reference for the tests
+const REF_GIT_TAG = 'benchmark';
```
pekam, line 19: What's the plan for this TODO?
Should the value be passed as a command-line argument?
Are you planning to resolve it within the scope of this PR?

tomivirkki: The plan was to have it updated once the PR was merged. But I guess we can just update the "benchmark" branch instead and not use a tag for this. Removed the TODO.

pekam: Ok, fine for me either way 👍

PR #980, opened 2020-05-15
Use the correct constructor for creating complete Person instances

### LINKED ISSUE #979: Demo: Grid filtering example not working as intended
Go to https://vaadin.com/components/vaadin-grid/java-examples/filtering and scroll down to "Using data type specific filtering"
The filter for "marital status" works on an attribute that isn't shown in the table. Also both options lead to an empty table.

image1

The birthdate attributes are not set thus filtering won't work either.

image2

tomivirkki:
Fixes #979

### DIFF
--- a/vaadin-grid-flow-demo/src/main/java/com/vaadin/flow/component/grid/demo/PersonData.java
+++ b/vaadin-grid-flow-demo/src/main/java/com/vaadin/flow/component/grid/demo/PersonData.java
@@ -1,7 +1,11 @@
 package com.vaadin.flow.component.grid.demo;
 
+import java.time.LocalDate;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Random;
+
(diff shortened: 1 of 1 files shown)


anasmi:
I guess we would also need to add into `GridView` around 1100 line 
>grid.addColumn(Person::getMaritalStatus).setHeader("Marital status");

to fix the issue closely? The grid still misses the `Marital status` column

tomivirkki:
@anasmi The demo design is really bad in many aspects (for example, with the birthday filter, you need to pass an exact date to filter by. You can't copy-paste values from the grid to the birthday filter because the date is not in the same format as in the date picker. The filters are difficult to clear because there's no clear filters button and the fields don't have clearButtonVisible either...). IMO the whole demo should be redesigned but that's another topic.

anasmi:
Yes, I agree with that.

Issue #974, opened 2020-05-13
Demo: ArrayIndexOutOfBoundsException on "Custom Drag Data" demo


Haprog:
### Description
Selecting a piece of text somewhere on the demo page and dragging it onto the grid in **Custom Drag Data** demo causes `ArrayIndexOutOfBoundsException`. This also happens if you drag an item from some of the other demos to this demo (at least when dragging an item from **Drag and Drop Filters** demo)

This causes errors to show up in our `#vaadincom-exceptions` Slack channel.

### Expected outcome
No exceptions should be thrown, no matter what you drag onto the grid (text, files, whatever).

### Actual outcome
An `ArrayIndexOutOfBoundsException` is thrown.

### Live Demo
https://vaadin.com/components/vaadin-grid/java-examples/drag-and-drop

### Steps to reproduce
1. Navigate to the **Custom Drag Data** demo at https://vaadin.com/components/vaadin-grid/java-examples/drag-and-drop
2. Select a piece of text for example from the source code block
3. Drag and drop the selected text onto the grid.

web-padawan:
Fixes by #983 which was merged to `4.2` branch.

PR #983, opened 2020-05-19
Fix ArrayIndexOutOfBoundsException in demo

anasmi:
Fixes https://github.com/vaadin/vaadin-grid-flow/issues/974 https://github.com/vaadin/90min-dx-tasks/issues/167

### DIFF
--- a/vaadin-grid-flow-demo/src/main/java/com/vaadin/flow/component/grid/demo/GridDemo.java
+++ b/vaadin-grid-flow-demo/src/main/java/com/vaadin/flow/component/grid/demo/GridDemo.java
@@ -20,9 +20,8 @@ import java.util.Set;
 import java.util.WeakHashMap;
 import java.util.stream.Collectors;
 
-import org.apache.commons.lang3.StringUtils;
-
 import com.vaadin.flow.component.ComponentEventListener;
+import com.vaadin.flow.component.Html;
 import com.vaadin.flow.component.button.Button;
 import com.vaadin.flow.component.checkbox.Checkbox;
 import com.vaadin.flow.component.combobox.ComboBox;
@@ -54,7 +53,6 @@ import com.vaadin.flow.component.grid.dnd.GridDropEvent;
 import com.vaadin.flow.component.grid.dnd.GridDropLocation;
 import com.vaadin.flow.component.grid.dnd.GridDropMode;
 import com.vaadin.flow.component.grid.editor.Editor;
-import com.vaadin.flow.component.Html;
 import com.vaadin.flow.component.html.Div;
 import com.vaadin.flow.component.html.H3;
 import com.vaadin.flow.component.html.Hr;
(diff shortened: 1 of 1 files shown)


vaadin-grid-flow-demo/src/main/java/com/vaadin/flow/component/grid/demo/GridDemo.java
```diff
@@ -2511,8 +2509,11 @@ private void createDragData() {
                         ? Collections.max(persons.stream().collect(Collectors
                                 .mapping(Person::getId, Collectors.toList())))
                         : 0;
-                persons.add(index, new Person(maxId + 1, dataArray[0],
-                        dataArray[1], -1, null, dataArray[2]));
+                if (dataArray.length == 3) {
```
shadikhani, line 2512: I think we can make it in a better way. For example, we can make this change in a way that it does not accept copied data from other demos. I suggest forcing this demo not to accept the data from other demos data by adding this code in addDropListener like this:

            if(event.getSource() != grid)
                return;

Now the code will look like this:
        grid.addDropListener(event -> {
            if(event.getSource() != grid)
                return;

            int index = event.getDropTargetItem().map(person -> persons
                    .indexOf(person)
                    + (event.getDropLocation() == GridDropLocation.BELOW ? 1
                            : 0))
                    .orElse(0);

            String personsData = event.getDataTransferText();
            Arrays.asList(personsData.split(";")).forEach(personData -> {
                String[] dataArray = personData.split(",");
                Integer maxId = !persons.isEmpty()
                        ? Collections.max(persons.stream().collect(Collectors
                                .mapping(Person::getId, Collectors.toList())))
                        : 0;
                persons.add(index, new Person(maxId + 1, dataArray[0],
                        dataArray[1], -1, null, dataArray[2]));
            });
            grid.getDataProvider().refreshAll();
        });

I think this would be easier to understand for our users.

anasmi: This is true and I think I tried it, but `event.getSource` returned grid for me anyway....

anasmi: Yes, unfrotuantely, it would not work as it always returns `Grid` (It returns the component whose Event it is, not the one who triggers it. Which is indeed misleading) 
https://vaadin.com/api/platform/14.2.0/com/vaadin/flow/component/grid/Grid.html#addDropListener-com.vaadin.flow.component.ComponentEventListener-

anasmi: @shadikhani

shadikhani: Yes, you are right!

PR #985, opened 2020-05-19
#912 Grid: always create one header row. Fixes #912

### LINKED ISSUE #912: Grid with no columns contains no header bar
This code will make the Grid to only have one header row:
```
        final Grid<Person> grid = new Grid<>(Person.class, false);
        final HeaderRow filterBar = grid.appendHeaderRow();
        final Grid.Column<Person> col = grid.addColumn("firstName");
        filterBar.getCell(col).setComponent(new TextField());
        add(grid);
```

Simply swapping two rows around will cause the Grid to have two header rows:
```
        final Grid<Person> grid = new Grid<>(Person.class, false);
        final Grid.Column<Person> col = grid.addColumn("firstName");
        final HeaderRow filterBar = grid.appendHeaderRow();
        filterBar.getCell(col).setComponent(new TextField());
        add(grid);
```

I believe this behaviour is inconsistent and both code snippets should cause the Grid to have two header rows.

mvysny:


tomivirkki:
We're currently in the process of moving all the component repositories to a centralized monorepo and need to close all open pull requests. If this is still needed, please re-apply to https://github.com/vaadin/vaadin-flow-components

PR #1002, opened 2020-05-28
Validate that size changes are fired.

caalador:
part of #8345


caalador:
@pleku @mshabarov I guess we will not put this in as the main of the test will be as an abstract in Flow?

mshabarov:
> @pleku @mshabarov I guess we will not put this in as the main of the test will be as an abstract in Flow?

Yes, I think so.

PR #1004, opened 2020-06-01
Use the size event to update label

caalador:
Use the size event to update
item set size.

### DIFF
--- a/vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/it/GridListDataViewPage.java
+++ b/vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/it/GridListDataViewPage.java
@@ -87,15 +87,13 @@ public class GridListDataViewPage extends Div {
                     dataView.withFilter(
                             item -> item.getFirstName().toLowerCase()
                                     .contains(event.getValue().toLowerCase()));
-                    // TODO: this is not needed after size change events are fired.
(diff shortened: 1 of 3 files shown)


vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/it/GridListDataViewPage.java
```diff
@@ -87,15 +87,13 @@ public GridListDataViewPage() {
...
-                    // TODO: this is not needed after size change events are fired.
-                    count.setText(Integer.toString(dataView.getDataSize()));
                 });
         filterByFirstName.setId(FIRST_NAME_FILTER);
 
         dataView.addSizeChangeListener(event -> {
             count.setText(Integer.toString(event.getSize()));
             showPreviousData.setEnabled(rowSelect.getValue() > 0);
-            showNextData.setEnabled(rowSelect.getValue() < event.getSize());
+            showNextData.setEnabled(rowSelect.getValue() < event.getSize() - 1);
```
pleku, line 96: I think there was API added for `HasNextItem` and `HasPreviousItem` that makes this even cleaner ?

caalador: Only for item not for row.
Perhaps we should add it for index also

PR #1029, opened 2020-06-12
Use setDataSource for setting data.

caalador:
part of vaadin/flow#8189

### DIFF
--- a/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
+++ b/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
@@ -18,6 +18,7 @@ package com.vaadin.flow.component.grid;
 import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashMap;
@@ -80,6 +81,7 @@ import com.vaadin.flow.data.provider.DataGenerator;
 import com.vaadin.flow.data.provider.DataProvider;
 import com.vaadin.flow.data.provider.DataProviderListener;
 import com.vaadin.flow.data.provider.HasDataGenerators;
+import com.vaadin.flow.data.provider.HasLazyDataView;
 import com.vaadin.flow.data.provider.HasListDataView;
 import com.vaadin.flow.data.provider.KeyMapper;
 import com.vaadin.flow.data.provider.ListDataProvider;
@@ -2299,7 +2301,13 @@ public class Grid<T> extends Component
         return -1;
     }
 
+    /**
+     * @inheritDocs
+     * @deprecated use instead one of the setDataSource methods from
+     * {@link HasListDataView} or {@link HasLazyDataView}
+     */
     @Override
+    @Deprecated
     public void setDataProvider(DataProvider<T, ?> dataProvider) {
         Objects.requireNonNull(dataProvider, "data provider cannot be null");
         handleDataProviderChange(dataProvider);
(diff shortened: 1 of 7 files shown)


vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
```diff
@@ -2299,7 +2301,13 @@ private int getLastFooterLayerIndex() {
         return -1;
     }
 
+    /**
+     * @inheritDocs
+     * @deprecated use instead one of the setDataSource methods from
```
pleku, line 2306:
```suggestion
     * @deprecated use instead one of the {@code setDataSource} methods which
```

vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
```diff
@@ -2299,7 +2301,13 @@ private int getLastFooterLayerIndex() {
...
+     * {@link HasListDataView} or {@link HasLazyDataView}
```
pleku, line 2307:
```suggestion
     * provide access to either {@link GridListDataView} or
     * {@link GridLazyDataView}..
```

vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
```diff
@@ -2299,7 +2301,13 @@ private int getLastFooterLayerIndex() {
...
+     */
     @Override
+    @Deprecated
     public void setDataProvider(DataProvider<T, ?> dataProvider) {
         Objects.requireNonNull(dataProvider, "data provider cannot be null");
```
pleku, line 2312: Should we move the implementation to a `setDataSource` method that is not deprecated and call that one from here ?

vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
caalador: Done, but can't link to GridLazyDataView here as it doesn't exist.

vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
caalador: done

vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
caalador: Perhaps not a setDataSource method, but perhaps some new private method that should be used.

vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/dataview/GridDataViewImpl.java
```diff
@@ -0,0 +1,79 @@
...
+
+    public GridDataViewImpl(DataCommunicator<T> dataCommunicator,
+            Grid<T> grid) {
+        this.dataCommunicator = dataCommunicator;
+        this.grid = grid;
+    }
+
+    @Override
+    public Stream<T> getItems() {
+        return dataCommunicator.getDataProvider()
```
pleku, line 43: Need to review the javadocs for this method separately as the implications for calling this method are very different for each three implementations of it

vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/dataview/GridDataViewImpl.java
```diff
@@ -0,0 +1,79 @@
...
+                .fetch(dataCommunicator.buildQuery(0, Integer.MAX_VALUE));
```
pleku, line 44: Not sure if this should use the `getSize()` as the limit instead of `Integer.MAX_VALUE`.
The lazy case could then override this as it probably should do anyway for the javadocs.

PR #1056, opened 2020-06-30
Merge support for unknown item count to master

pleku:
Depends on flow part https://github.com/vaadin/flow/pull/8657


pom.xml
```diff
@@ -17,6 +17,7 @@
 
     <properties>
         <contextmenu.version>4.2-SNAPSHOT</contextmenu.version>
+        <flow.version>3.2-SNAPSHOT</flow.version>
```
mshabarov, line 20: Has to be deleted because the flow version is in `flow-component-base` pom

mshabarov, line 20:
```suggestion
```

pleku:
Merged before the snapshot was available, will retrigger a new build when it is

PR #1064, opened 2020-07-05
Add getItem(int index) to DataView #8652

mshabarov:
Part of #8676
Flow part https://github.com/vaadin/flow/pull/8676


mshabarov:
This PR was combined with https://github.com/vaadin/vaadin-grid-flow/pull/1063 because the corresponding changes in Flow depends to each other, and otherwise the builds will failing due to compile errors

Issue #1082, opened 2020-07-29
Empty Grid gets stuck in loading state


tomivirkki:
In the following minimalistic example, the empty grid is added on the page, but the `<vaadin-grid loading>` Web Component will have the `loading` property/attribute on until data gets added to the grid.
```java
TreeGrid<String> grid = new TreeGrid<>();
add(grid);
```

Expected result: an empty Grid should not have the `loading` attribute. 

This is because the `rootPageCallbacks[page]` callback is never executed unless [this](https://github.com/vaadin/vaadin-grid-flow/blob/dc3e17e2a5d45982955128ab12102bfe814830e8/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js#L349) condition is true but it's left pending indefinitely instead.

Issue #802, opened 2019-10-25
Unsynchronized "expand" in TreeGrid


ChristianKerth-IOSB:
The expansion of TreeGrids seems to be an asynchr. event sent to the client. When UI elements in the TreeGrid are updated (e.g. in a ComponentColumn that has icons) while the TreeGrid is still expanding on client-side, there are exceptions.

How to reproduce:
- have a "large" TreeGrid (~50 nodes, 4-5 levels deep)
- show the TreeGrid in a client-view, expand entire TreeGrid on server side
- update the entries in the TreeGrid, so that there are changes in the UI (while the client still expands the TreeGrid)

Results: 
- it takes several seconds until the TreeGrid expands
- Exceptions on client side, sometimes an empty TreeGrid is shown, sometimes the ComponentValues are not rendered any more, sometimes the server-side receives an exception that "node IDs are unknown on client side".

Expected Result:
- the TreeGrid expands within a really short time (well below a second)
- server-side and client-side are synchronized in a way, that the client side already knows about IDs when it receives instructions from the server-side to update them.

Version used: Vaadin 14.0.5 (via vaadin-spring-boot-starter)

TatuLund:
See also https://github.com/vaadin/vaadin-grid-flow/issues/476

tulioag:
This example reproduces the performance issues.
[treegriddemo.zip](https://github.com/vaadin/vaadin-grid-flow/files/4408658/treegriddemo.zip)

pekam:
I'm not able to reproduce the issue.

I'm using `vaadin-grid-flow` 4.0.6, which belongs to Vaadin 14.0.5.

With the following code, I click the first button to expand all items. While the grid is expanding, I click the second button which refreshes both the first item (already rendered), and the last item (still waiting to be loaded). Everything works as expected; there are no exceptions, the first item is updated immediately and the last one renders the updated value when loaded. Using `refreshAll` instead of `refreshItem` works as well.
```java
@Route("treegrid-deep-tree")
public class TreeGridDeepTreePage extends Div {

    private static final int DEPTH = 50;

    private TreeGrid<Person> treeGrid;
    private List<Person> items;
    private TreeDataProvider<Person> dataProvider;

    public TreeGridDeepTreePage() {
        initializeDataProvider();
        treeGrid = new TreeGrid<>();
        treeGrid.setDataProvider(dataProvider);
        treeGrid.setWidth("100%");
[...]

PR #1084, opened 2020-08-07
Perf/unnecessary page calls

### LINKED ISSUE #1040: Massive data requests when scrolling quickly
When you scroll quickly through a Grid (i.e. by dragging the scrollbar), Grid requests data for lots of intermediate rows and not only the rows where the user ends up.

Can repreproduced with the example application from https://github.com/vaadin/vaadin-grid-flow/issues/765#issuecomment-643958986. From one erratic but relatively constant drag of the scrollbar, I observed two primary round trips. The first contained three separate `setRequestedRange` invocations that each requested data for 150 rows starting at indices 450, 800 and 950. The resulting response was 2.2MB gzipped for a total of 4.5MB of JSON. Processing all this data took at total of ~4 seconds. Immediately when this was completed, a new request is sent with the invocations that have been enqueued while the previous one was ongoing. This one contains one `confirmUpdate` invocation and 6 more `setRequestedRange` (150 rows each) starting at indices 1000, 1100, 1250, 1300, 1450 and 1500. The resulting response is slightly bigger (5.1MB of JSON) and takes around 4.5 seconds to process.

I suspect some kind of debounce would be necessary before requesting rows. A regular debounce might not be appropriate since we would still not want to prevent regular prefetching when the user slowly scrolls through the data.

tomivirkki:
Fixes #1040 
Fixes https://github.com/vaadin/vaadin-grid-flow/issues/1082
Relates to #802


vaadin-grid-flow-testbench/src/main/java/com/vaadin/flow/component/grid/testbench/GridElement.java
```diff
@@ -32,6 +32,18 @@
...
+    protected void waitUntilLoadingFinished() {
+        waitUntil(e -> !isLoading());
+    }
+
+    protected boolean isLoading() {
+        return (Boolean) executeScript("return arguments[0]._cache.isLoading()",
+                this)
+                || (Boolean) executeScript(
+                        "return arguments[0].$connector.hasRootRequestQueue();",
+                        this);
```
DiegoCardoso, line 44: I wonder if this can be done within a single `executeScript` call and if it's worth changing it.

DiegoCardoso (commented):
Changes at the connector looks good to me. Checked the tests and looks good too. Only one small comment, that is not a blocker for me.

vaadin-grid-flow-testbench/src/main/java/com/vaadin/flow/component/grid/testbench/GridElement.java
tomivirkki: Good point. Fixed in both GridElement and TreeGridElement.

DiegoCardoso (approved):

DiegoCardoso:
I tested it in Bakery and didn't find any issue while running the project.

PR #1100, opened 2020-08-20
Filter support in Data View API

mshabarov:
These changes in Grid tied with this Data View API updates in Flow https://github.com/vaadin/flow/pull/8874

### DIFF
--- a/vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/it/dataview/AbstractItemCountGridPage.java
+++ b/vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/it/dataview/AbstractItemCountGridPage.java
@@ -218,15 +218,8 @@ public abstract class AbstractItemCountGridPage extends VerticalLayout
     }
 
     protected void switchToDefinedSize() {
-        grid.getLazyDataView().setItemCountCallback(query -> {
-            // LazyLoadingProvider has no filter, and this callback is given
-            // a filter type of Object. Thus, this ugly Object->Void conversion
-            // shall be done.
-            Query<String, Void> noFilterQuery = new Query<>(query.getOffset(),
-                    query.getLimit(), query.getSortOrders(),
-                    query.getInMemorySorting(), null);
-            return dataProvider.size(noFilterQuery);
-        });
+        grid.getLazyDataView()
+                .setItemCountCallback(query -> dataProvider.size(query));
         dataProviderSizeInput.setEnabled(true);
     }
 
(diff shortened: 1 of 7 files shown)


vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/it/dataview/AbstractItemCountGridPage.java
```diff
@@ -218,7 +218,15 @@ protected void switchToDataProvider() {
     }
 
     protected void switchToDefinedSize() {
-        grid.getLazyDataView().setItemCountCallback(dataProvider::size);
+        grid.getLazyDataView().setItemCountCallback(query -> {
```
pleku, line 221: Another option could have been to change the data provider filter to be `Object` instead of `Void` ?

Now I realize the hindrance, luckily the only one, of not having the filter type parameter in `LazyDataView`. It makes this `itemCountCallback` method a bit useless for both `Grid` and for `ComboBox`. 

Since keeping it makes sense for switching between defined/undefined size without a full reset, I wonder if we should just implement it in the concrete LazyDataView classes for the components, so that the type could go right. It would have some code duplication, but then again the type would be immediately the correct one for the users.

vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/it/dataview/AbstractItemCountGridPage.java
```diff
@@ -218,7 +218,8 @@ protected void switchToDataProvider() {
     }
 
     protected void switchToDefinedSize() {
-        grid.getLazyDataView().setItemCountCallback(dataProvider::size);
+        grid.getLazyDataView()
+                .setItemCountCallback(query -> dataProvider.size(query));
```
pleku, line 222: Use method reference instead, meaning revert unnecessary change

pleku (approved):

vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/it/dataview/AbstractItemCountGridPage.java
mshabarov: Hmm, why did I do this...Reverted

pleku (approved):

pleku:
Now there is a new flow snapshot so the build should be green.

mshabarov:
Components uses `4.0-SNAPSHOT` and master has `4.1-SNAPSHOT`, so let's do the cherry-pick to 4.0

PR #1108, opened 2020-09-03
Connector cleanup

tomivirkki:
...for example, the internal flag `ensureSubCacheDelay` is fixed to `true`, and I don't see why this constant would be changed anytime soon in the future, so it's pointless to have logic in the connector that only applied if the flag was `false`.


DiegoCardoso (approved):
LGTM

PR #1111, opened 2020-09-04
fix: prevent NPE for client-side only grids

### LINKED ISSUE #1077: Adding a tree grid from client side and another grid from server side causes javascript error
When we add a tree grid for navigation purpose in our layout js module and then afterwards add another grid from java on the same view, we get following error:
_Cannot read property 'beforeEnsureSubCacheForScaledIndex' of undefined_
Also the tree grid will immediately collapse and not expand anymore, but the toggle icons will change correctly.

Adding both grids from only client or server side works. But in our case we would definitely need the mixed variant.

It can be easily reproduced by adding a js module with a tree grid and a slot for other components inside and creating related java class with an add method, in order to add elements to that slot. When the view is attached, just add another grid to that slot and also provide data.

DiegoCardoso:
fix #1077

`ItemCache#ensureSubCacheForScaledIndex` method is changed by `gridConnector`, so it will make any grid on the page use the new implementation. That leads to a call from undefined object in case of a grid entirely defined on the client side needs to call this method.

This change stores the original implementation and calls it when the grid doesn't have a `$connector` object.

### DIFF
--- a/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
+++ b/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
@@ -15,7 +15,15 @@ import { ItemCache } from '@vaadin/vaadin-grid/src/vaadin-grid-data-provider-mix
         return;
       }
 
+      // Storing original implementation of the method to be used for client
+      // side only grids
+      ItemCache.prototype.ensureSubCacheForScaledIndexOriginal = ItemCache.prototype.ensureSubCacheForScaledIndex;
       ItemCache.prototype.ensureSubCacheForScaledIndex = tryCatchWrapper(function(scaledIndex) {
+        if (!this.$connector) {
+          this.ensureSubCacheForScaledIndexOriginal(scaledIndex);
+          return;
+        }
+
         if (!this.itemCaches[scaledIndex]) {
 
           if(ensureSubCacheDelay) {
--- a/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
+++ b/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
@@ -19,7 +19,7 @@ import { ItemCache } from '@vaadin/vaadin-grid/src/vaadin-grid-data-provider-mix
       // side only grids
       ItemCache.prototype.ensureSubCacheForScaledIndexOriginal = ItemCache.prototype.ensureSubCacheForScaledIndex;
       ItemCache.prototype.ensureSubCacheForScaledIndex = tryCatchWrapper(function(scaledIndex) {
-        if (!this.$connector) {
+        if (!this.grid.$connector) {
           this.ensureSubCacheForScaledIndexOriginal(scaledIndex);
           return;
         }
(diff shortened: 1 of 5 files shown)


vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
```diff
@@ -15,7 +15,15 @@ import { ItemCache } from '@vaadin/vaadin-grid/src/vaadin-grid-data-provider-mix
         return;
       }
 
+      // Storing original implementation of the method to be used for client
+      // side only grids
+      ItemCache.prototype.ensureSubCacheForScaledIndexOriginal = ItemCache.prototype.ensureSubCacheForScaledIndex;
```
DiegoCardoso, line 20: Wasn't sure how to call the stored method, suggestions are welcome.

vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
```diff
@@ -15,7 +15,15 @@ import { ItemCache } from '@vaadin/vaadin-grid/src/vaadin-grid-data-provider-mix
...
       ItemCache.prototype.ensureSubCacheForScaledIndex = tryCatchWrapper(function(scaledIndex) {
```
DiegoCardoso, line 21: Also noticed that these changes to `ItemCache` prototype are done on every grid added to the page. I thought of adding a check for it to be initialized only once (but keeping it along aside the first grid initialization). I wasn't sure if that would have any impact (good or bad) to the component, so I didn't add, but if we think that this would be beneficial, we could bring this change in another PR.

tomivirkki: Sounds like a good idea. Maybe it would work if the prototype patching was moved outside of the `initLazy` function.

vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
tomivirkki: You mean if you stored it as `const ensureSubCacheForScaledIndexOriginal  = ...`?

Would it work if it was then called with `ensureSubCacheForScaledIndexOriginal.call(this, scaledIndex)`?

DiegoCardoso: My bad. I meant how to name the method. Not sure if "Original" at the end is good.

> Would it work if it was then called with ensureSubCacheForScaledIndexOriginal.call(this, scaledIndex)?

I thought of calling the method this way, but since the prototype patching is being done everywhere, I decided to go on the same path.

vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
DiegoCardoso: The idea of keeping it inside of `initLazy` was to delay the patching to the moment where it's needed (once a server-side grid is first attached). But I believe that would be the majority of cases, so it might not be necessary to postpone the prototype patching.

tomivirkki (approved):

